PR #13, opened 2025-06-02
Tags plugin

SonjaPopovic:
https://jira.corp.adobe.com/browse/MWPW-173142

Copy of tags plugin from da-bacom-blog, with update:
- to expect existing json structure of config (json.data.data instead of json.data)
- default behaviour stays the same as for bacom-blog
- supported additional configs to support fetching from AEM author instances for projects that don't have aem cloud

### DIFF
--- a/nx/public/plugins/tags.html
+++ b/nx/public/plugins/tags.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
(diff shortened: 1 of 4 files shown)


SonjaPopovic:
There is a readme step - use npm run commit for commits, but there are is no commit script in package.json, is this still relevant?

Issue #16, opened 2025-06-02
Adding Permission Fails for Roles Without Users


Ben-Zahler:
on https://da.live/apps/aem-permissions#, if none of the users have the role that a new user is requesting, there is a JS error that prevents the permissions to be applied.

<img width="1259" alt="Image" src="https://github.com/user-attachments/assets/2cc5bf49-b32d-4206-8c75-9bfa5f40bb72" />

utils.js:204 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'some')
    at utils.js:204:40
    at Array.forEach (<anonymous>)
    at approveSiteUser (utils.js:196:13)
    at async NxPermissions.handleUserAction (

PR #17, opened 2025-06-02
fix: #16 initialize roles without users

Ben-Zahler:
Initialize all roles to empty array if they don't exist.

Fix https://github.com/adobe/da-nx/issues/16


auniverseaway:
We don't want to persist an empty array into Helix if there are no users. It's a bit noisy.

Are you facing a bug that led you down this path?

auniverseaway:
I believe this has been solved another way. @Ben-Zahler if I'm mistaken, let's discuss on Slack.

PR #29, opened 2025-06-20
Add disabled to sl-button, focus to sl-input

chrischrischris:


nx/public/sl/components.js
```diff
@@ -229,6 +236,7 @@ class SlButton extends LitElement {
       <span class="sl-button">
         <button
           class="${this.class}"
+          ?disabled="${this.disabled}"
```
auniverseaway, line 239: `${spread(this._attrs)}` already should give us this.

auniverseaway (approved):

PR #62, opened 2025-08-18
fix: enhance URL parsing in shell.js to include search and hash

mhaack:
Shell for DA Apps should pass through URL hash's as well.


sirugh (approved):
LGTM. One nit.

auniverseaway (approved):

PR #81, opened 2025-09-09
feat: upgrade to rum v2

kptdobe:
Upgrade to RUM v2.

We'll be tracking clicks in webcomponents and many other good things like better source handling.

Not sure how to provide a URL for testing...

### DIFF
--- a/nx/scripts/nexter.js
+++ b/nx/scripts/nexter.js
@@ -114,7 +114,7 @@ function decorateLinks(el) {
 
 function decorateSections(parent, isDoc) {
   const selector = isDoc ? 'main > div' : ':scope > div';
-  return [...parent.querySelectorAll(selector)].map((el, i) => {
+  return [...parent.querySelectorAll(selector)].map((el) => {
     el.classList.add('section');
     el.dataset.status = 'decorated';
     el.autoBlocks = decorateLinks(el);
--- a/nx/scripts/nexter.js
+++ b/nx/scripts/nexter.js
@@ -1,5 +1,3 @@
-import { sampleRUM } from '../deps/rum.js';
-
 const AUTO_BLOCKS = [
   { 'nx-fragment': '/fragments/' },
   { 'nx-youtube': 'https://www.youtube.com' },
@@ -122,9 +120,6 @@ function decorateSections(parent, isDoc) {
     el.autoBlocks = decorateLinks(el);
     el.blocks = [...el.querySelectorAll(':scope > div[class]')];
     decorateDefaults(el);
-    if (i === 0 && sampleRUM.enhance) {
-      sampleRUM.enhance();
-    }
     return el;
   });
 }
--- a/nx/scripts/nexter.js
+++ b/nx/scripts/nexter.js
@@ -1,9 +1,9 @@
+import { sampleRUM } from '../deps/rum.js';
+
 const AUTO_BLOCKS = [
   { 'nx-fragment': '/fragments/' },
   { 'nx-youtube': 'https://www.youtube.com' },
 ];
-
-// Hello World
 
 function getEnv() {
   const { host } = new URL(window.location.href);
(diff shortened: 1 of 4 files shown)


nx/scripts/nexter.js
```diff
@@ -116,12 +116,15 @@ function decorateLinks(el) {
...
 function decorateSections(parent, isDoc) {
   const selector = isDoc ? 'main > div' : ':scope > div';
-  return [...parent.querySelectorAll(selector)].map((el) => {
+  return [...parent.querySelectorAll(selector)].map((el, i) => {
     el.classList.add('section');
     el.dataset.status = 'decorated';
     el.autoBlocks = decorateLinks(el);
     el.blocks = [...el.querySelectorAll(':scope > div[class]')];
     decorateDefaults(el);
+    if (i === 0 && sampleRUM.enhance) {
```
auniverseaway, line 125: What does `sampleRUM.enhance` do? We should not have a dependency on rum for basic decorating. It provides the user trying to access the page no value.

nx/scripts/postlcp.js
```diff
@@ -1,6 +1,10 @@
 import { loadBlock } from './nexter.js';
+import { sampleRUM } from '../deps/rum.js';
 
 (async function loadPostLCP() {
+  window.hlx = window.hlx || {};
```
auniverseaway, line 5: These should be moved to `rum.js` as they are part of the rum feature. We want to keep these things encapsulated to their own area of concern.

auniverseaway (changes requested):
I've always had a complicated relationship with RUM because it provides the site owner value, it provides Adobe value, but it does not provide any value to the visitor.

In simple Edge Delivery implementations with a lot of LCP headroom, it's totally fine to put RUM in front of LCP. That's the way boilerplate is. This is not something I'm willing to accept on DA.

If we cannot get the RUM data we need without putting bytes in front of LCP, I think we have an architectural problem with RUM. Historically, we have been able to defer without any noticeable impacts to reporting (at least to my knowledge), and I don't want to change this.

If you look at it differently: if AWS / Cloudflare / etc said we had to run their code before we can do our work, we would tell them to fly a kite. We should treat our customers (da.live in this case) with the same amount of respect.

kptdobe:
> but it does not provide any value to the visitor.

It does provide value to the visitor: if we can track what happens when a page loads, we can detect when page is slow or if something is broken and then, we can work hard to make it better to the visitor.

> if AWS / Cloudflare / etc said we had to run their code before we can do our work

Yes and this is the case for almost all projects. Then it is a question of tradeoff: do you accept to lose x% of performance for the benefits of the given feature... 

The `sampleRUM` function is adding some bytes in front of the LCP. That's true and nothing we can do about it because we want to capture things as early possible (see tradeoff value for visitors above). It has a measured impact of 15 to 30ms. For a LCP at 2800ms, this is irrelevant and largely in the error margin.
When we talk with Akamai that is adding scripts in the way to LCP, we are talking about 2 to 3... seconds!

Is another site than da.live using this repo as a foundation ? Because in the current state and for da.live, the LCP argument is irrelevant.
- for the explorer, you have to load the list of folder' children, that's all what really matter. The LCP event comes really late anyway, just because there is nothing really "large". 95 resources are loaded because the LCP, you can move the rum.js resource wherever you want, it will not change anything
image

- for the editor, the LCP will most likely be the hero image of the page being edited. It is the last thing loaded on the page (by nature of an editor). If we want to fine tune, we should question why the whole da-library is loaded before.
image

Finally, by deferring the load of rum.js, you lose visibility of everything that happens before (early checkpoints, resource loaded (or not), potential JS errors...).

nx/scripts/nexter.js
kptdobe: `enhance` is an optimization to load the enhancer only after the first section is loaded, i.e. load the enhancer post LCP.

kptdobe: If we lazy load `sampleRUM`, this would not be necessary.

auniverseaway (approved):

PR #97, opened 2025-10-05
Snapshot scheduler

### LINKED ISSUE #96: Add Scheduled Publish to DA Snapshot UI
Add the new helix-snapshot-scheduler service to be triggered from the DA Snapshot experience. It is already connected to the Snapshot Admin in labs.aem.live but needs to have the same integration in DA as well.

amol-anand:
Fix #96 

Test URLs:
- Before: https://main--da-nx--adobe.aem.page/drafts/amol/scheduler
- After: https://snapshot-scheduler--da-nx--amol-anand.aem.page/drafts/amol/scheduler

### DIFF
--- a/nx/blocks/snapshot-admin/views/snapshot.js
+++ b/nx/blocks/snapshot-admin/views/snapshot.js
@@ -8,7 +8,6 @@ import {
   copyManifest,
   updatePaths,
   reviewSnapshot,
-  isRegistered,
   updateSchedule,
   formatLocalDate,
 } from '../utils/utils.js';
@@ -31,19 +30,18 @@ const ICONS = [
 class NxSnapshot extends LitElement {
   static properties = {
     basics: { attribute: false },
+    isRegistered: { attribute: false },
     _manifest: { state: true },
     _editUrls: { state: true },
     _message: { state: true },
     _isOpen: { state: true },
     _action: { state: true },
-    _isRegistered: { state: true },
   };
 
   async connectedCallback() {
     super.connectedCallback();
     this.shadowRoot.adoptedStyleSheets = [style];
     getSvg({ parent: this.shadowRoot, paths: ICONS });
-    this._isRegistered = await isRegistered();
   }
 
   update(props) {
@@ -284,7 +282,7 @@ class NxSnapshot extends LitElement {
             <sl-textarea name="description" resize="none" .value="${this._manifest?.description}"></sl-textarea>
             <p class="nx-snapshot-sub-heading">Password</p>
             <sl-input type="password" name="password" .value=${this._manifest?.metadata?.reviewPassword}></sl-input>
-            ${this._isRegistered ? html`
+            ${this.isRegistered ? html`
               <p class="nx-snapshot-sub-heading">Schedule Publish</p>
               <sl-input type="datetime-local" name="scheduler" .value=${formatLocalDate(this._manifest?.metadata?.scheduledPublish)}></sl-input>
             ` : nothing}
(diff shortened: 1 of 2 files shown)


nx/blocks/snapshot-admin/utils/utils.js
```diff
@@ -1,4 +1,4 @@
-import { AEM_ORIGIN } from '../../../public/utils/constants.js';
+import { AEM_ORIGIN, SNAPSHOT_SCHEDULER_URL } from '../../../public/utils/constants.js';
```
auniverseaway, line 1: If nothing else is using this, I would keep this internal to this file.

nx/blocks/snapshot-admin/utils/utils.js
```diff
@@ -165,3 +165,38 @@ export async function copyManifest(name, resources, direction) {
...
+    return false;
+  }
+}
+
+// Convert UTC date to local datetime-local format
+export function formatLocalDate(utcDate) {
+  if (!utcDate) return '';
+  const d = new Date(utcDate);
+  return `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, '0')}-${String(d.getDate()).padStart(2, '0')}T${String(d.getHours()).padStart(2, '0')}:${String(d.getMinutes()).padStart(2, '0')}`;
+}
```
auniverseaway, line 202: We have functions that already do this in our preferred formats that are also international friendly.

```
export function formatDate(timestamp) {
  const rawDate = timestamp ? new Date(timestamp) : new Date();
  const date = rawDate.toLocaleDateString([], { year: 'numeric', month: 'short', day: 'numeric' });
  const time = rawDate.toLocaleTimeString([], { hour: 'numeric', minute: '2-digit' });
  return { date, time };
}
```

nx/blocks/snapshot-admin/views/snapshot.js
```diff
@@ -33,14 +36,21 @@ class NxSnapshot extends LitElement {
...
   async connectedCallback() {
     super.connectedCallback();
+    this._isRegisteredForSnapshotScheduler = false;
     this.shadowRoot.adoptedStyleSheets = [style];
     getSvg({ parent: this.shadowRoot, paths: ICONS });
   }
 
+  async firstUpdated() {
+    // Check if registered for snapshot scheduler after first render
+    this._isRegisteredForSnapshotScheduler = await isRegisteredForSnapshotScheduler();
```
auniverseaway, line 51: If you're not waiting on DOM to render, you can do this in `connectedCallback` which should only run once.

`firstUpdated` is for use cases where you need to know the DOM inside the web component.

https://lit.dev/docs/components/lifecycle/#firstupdated (this whole page on lifecycle is great, btw)

nx/blocks/snapshot-admin/views/snapshot.js
```diff
@@ -83,6 +93,23 @@ class NxSnapshot extends LitElement {
     // Set the name if it isn't already set
     if (!this.basics.name) this.basics.name = name;
 
+    // Validate scheduled publish time before saving
+    const scheduledPublish = this.getValue('[name="scheduler"]');
+    if (scheduledPublish) {
```
auniverseaway, line 98: We like early returns.

auniverseaway (changes requested):
Minor things. When you address we can pull it into a scheduler branch so we can run this on some production-ish snapshots.

Thanks for doing this. You may not realize it, but this unlocks a few things we want to do on the security front for DA.

PR #127, opened 2025-11-25
nx-shell user warning

hannessolo:
Warn the user that they're giving their token to the app loaded by nx-shell.


hannessolo:
> I wonder if we also do this based on ref

That's exactly what it's doing :) 

Screenshot 2025-11-25 at 19 05 09

PR #143, opened 2025-12-17
Fix issue #142 - Sandbox toast bug

### LINKED ISSUE #142: getIsSandbox not handling 403 or 401 errors correctly
`if (status === 403 && status === 401) return false;` should be `if (status === 403 || status === 401) return false;`

JordanRhode:
Fix #142

Description: this bug causes the sandbox toast to show incorrectly when someone does not have authorization to read the DA config.
image

### DIFF
--- a/nx/utils/org-check.js
+++ b/nx/utils/org-check.js
@@ -10,7 +10,7 @@ async function getIsSandbox(org) {
   const { status } = confResp;
 
   // Handle not authorized
-  if (status === 403 && status === 401) return false;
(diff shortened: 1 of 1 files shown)


JordanRhode:
CLA bot didn't detect my signature. Opening new PR to try again.

auniverseaway:
@JordanRhode thanks for this. 

**Note:** if you want someone to use a DA project, they should at least have read access to config. Without read, they will not be access to many configurations: library, AEM Asset configs, UE prefs, etc.

PR #149, opened 2026-01-09
Remove all diff tags before doing diffs

chrischrischris:
Also:
* moves Launch commands to own section on right panel
* adds tooltips for sync and promote links
* input validation of snapshot name

Test URLs:
- Before: https://da.live/apps/snapshots#/adobecom/da-bacom
- After: https://main--da-live--adobe.aem.live/apps/snapshots?nx=snapdif#/adobecom/da-bacom


auniverseaway (dismissed):
Small nit. Otherwise, looks great. Much clearer.

auniverseaway (approved):

PR #185, opened 2026-02-10
MWPW-187566 – English copy files missing in destination

raga-adbe-gh:
While performing the English copy, the langPath is not being populated hence replacing this with daDestPath.

Fix #MWPW-187566


auniverseaway (approved):
This will likely break some other stuff, but I don't think anyone is using it.

PR #189, opened 2026-02-12
Translation Metadata addition to GLaaS

janaki-r-bhagwath:
- HTML annotation for maxlength and keywords based on block-schema
(its-storage-size, its-loc-note, its-loc-note-type)
- Per-language keywords metadata (langMetadata) sent to GLaaS
- Single `<p>` tag unwrapping for cleaner HTML structure

### DIFF
--- a/nx/blocks/loc/connectors/glaas/translationMetadata.js
+++ b/nx/blocks/loc/connectors/glaas/translationMetadata.js
@@ -3,7 +3,7 @@ import { daFetch } from '../../../../utils/daFetch.js';
 
 const BLOCK_SCHEMA_PATH = '/.da/block-schema.json';
 
-let BLOCK_SCHEMA_CACHE;
+let blockSchemaCache;
 
 export function processSchemaKey(schemaKey) {
   const match = schemaKey.match(/^([\w-]+)\s*\((.*)\)$/);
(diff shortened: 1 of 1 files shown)


nx/blocks/loc/connectors/glaas/translationMetadata.js
```diff
@@ -0,0 +1,275 @@
...
+  const parsedSchema = {};
+
+  Object.keys(schemaData).forEach((key) => {
+    if (key.startsWith(':')) return;
+    const blockData = schemaData[key];
+    if (!blockData.data) return;
+    const { id, selector } = processSchemaKey(key);
+    const fields = [];
+    blockData.data.forEach((field) => {
+      const fieldName = field['field name'];
```
chrischrischris: No idea how rigid schemaData is, but is there any need to possibly support `field_name` too?

nx/blocks/loc/connectors/glaas/translationMetadata.js
```diff
@@ -0,0 +1,275 @@
+import { DA_ORIGIN } from '../../../../public/utils/constants.js';
+import { daFetch } from '../../../../utils/daFetch.js';
+
+const BLOCK_SCHEMA_PATH = '/.da/block-schema.json';
+
+let BLOCK_SCHEMA_CACHE;
```
chrischrischris: nit: All caps usually used for constants

chrischrischris (commented):
A few optimizations to reduce querySelector usage.  Might be overkill if the docs are small

Issue #184, opened 2026-02-10
[dasc] Field validation UI


auniverseaway:
As an author, I would like to know if my object, array, or simpleton is not valid.

## Criteria of acceptance
1. Simpleton's show errors below the field.
2. Objects highlight red (soft red background?)
3. Objects have human-readable text stating the issue
4. Arrays highlight red (soft red background?)
5. Arrays have a human-readable text stating the issue
6. Uses existing FormModel validation function.
7. Any form with errors should have _some kind_ of warning.

## Out of scope
1. Showing any errors in the outline - I'm not convinced this is needed. We can revisit.

PR #195, opened 2026-02-16
feat: Implement basic form validation.

kozmaadrian:
Ticket: https://github.com/adobe/da-nx/issues/184

https://github.com/user-attachments/assets/8a58feb9-89be-4587-ab7b-19a719a2e835

### DIFF
--- a/nx/blocks/form/data/model.js
+++ b/nx/blocks/form/data/model.js
@@ -2,8 +2,8 @@ import { DA_ORIGIN } from 'https://da.live/blocks/shared/constants.js';
 import { daFetch } from 'https://da.live/blocks/shared/utils.js';
 import HTMLConverter from '../utils/html2json.js';
 import JSONConverter from '../utils/json2html.js';
-import { Validator } from '../../../deps/da-form/dist/index.js';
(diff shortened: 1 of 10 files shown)


nx/blocks/form/data/model.js
```diff
@@ -55,20 +54,20 @@ export default class FormModel {
   }
 
   updateProperty({ name, value }) {
-    setValueByPath(this._json, name, value);
+    setValueByPointer(this._json, name, value);
```
kozmaadrian: Using the standard naming Pointer (as a best practice) instead of Path. The same renaming is applied later in the following files.

nx/blocks/form/utils/generator.js
```diff
@@ -74,7 +74,7 @@ export default function generateEmptyObject(
       return schema.enum && schema.enum.length > 0 ? schema.enum[0] : 0;
 
     case 'boolean':
-      return schema.enum && schema.enum.length > 0 ? schema.enum[0] : false;
+      return false;
```
kozmaadrian: Handles the boolean default value correctly.

Issue #196, opened 2026-02-17
[dasc] Persist Only Required Fields, Schema Defaults, and User Data


kozmaadrian:
## Summary

The form should save only schema dedined mandatory fields, schema defined defaults or user-entered data only.

---

## General Functionality

The current implementation uses the JSON Schema + JSON data to dynamically generate the editor UI. There are two cases:

- **Creating a new document**: In this case the data is auto-generated with default empty values.
- **Authoring an existing document**: In this case the editor will get already saved data loaded.

A field can have a value from one of:

- **Default from schema** — Value from the schema `default` keyword
- **User-entered** — Value typed or selected by the user
- **Auto-generated** — Value produced by the data generator (e.g. empty string, empty array) for UI rendering

---

## Problem

Currently the form stores the **complete** JSON data—not only the user-entered data but the auto-generated default values as well.

The generated values are used to ensure the UI can render the corresponding input with some default values. However, all these generated values are **not needed to be saved all the time**.

---

## Expected Behavior

JSON Schema supports the option to mark a field as [required](https://www.learnjsonschema.com/2020-12/validation/required/). This feature adds the restriction that the property marked required must be present in the JSON data, but does not enforce anything about its value.

The editor should store the JSON data having (in priority order):

1. **P1** — All data properties that are marked as required
2. **P2** — Default values defined by the schema
3. **P3** — If validation is defined in the schema (e.g. `format`, etc.) → only valid values are stored
4. **P4** — User-entered data — but not auto-generated default values

---

## Additional benefit: JSON to HTML Output

In this way, saving the JSON as HTML would also remove a lot of empty unnecessary blocks.

Issue #181, opened 2026-02-10
[dasc] Ability to add & remove array items


auniverseaway:
As an author, I would like the ability to add a new item to an array.

## Criteria of acceptance
1. Ability to add a new item in an array.
2. Add new uses `items.schema` properties to infer type.4. Ability to remove an item.
3. Supports both primitives and objects.
4. JSON preview updates
5. Outline view updates
6. Uses existing persistence handler to automatically persist changes
7. Adds to bottom of existing array.
8. Item is highlighted in a way that is consistent with other focused items

## Out of scope
1. Reordering
2. Copying / duplicating
3. Scrolling to position of new item (will be handled through an over-arching scrollTo handler)

PR #198, opened 2026-02-19
feat: dasc #181 handle arrays, #196 annotate based on schema

kozmaadrian:
While working on #181, we realised that we need to modify a core part of the annotation mechanism described in #196.

* Ticket: https://github.com/adobe/da-nx/issues/181
* Ticket: https://github.com/adobe/da-nx/issues/196

The refactored annotation mechanism depends on the changes introduced in #181 to properly test both the UI behavior and data persistence.

The updates are split across two separate PRs:
* Ticket: 181 -> PR https://github.com/adobe/da-nx/pull/190
* Ticket: 196 -> PR https://github.com/adobe/da-nx/pull/197 (built on top of PR [190](https://github.com/adobe/da-nx/pull/190)

The current PR is intended to provide a clearer view of all changes compared to the main branch.

Demo:

https://github.com/user-attachments/assets/9104d800-07bf-4d42-9da2-8181089d2aee

### DIFF
--- a/nx/blocks/form/views/editor.css
+++ b/nx/blocks/form/views/editor.css
@@ -17,6 +17,10 @@
   }
 }
 
+.primitive-item-content {
+  margin-block: 16px 4px;
+}
+
 .root-group.item-group {
   > .item-group-title {
     > p {
@@ -80,10 +84,6 @@
       order: 2;
     }
   }
-}
-
-p.primitive-item-title {
-  margin-block: 16px 4px;
 }
 
 .add-item-btn {
--- a/nx/blocks/form/views/editor.js
+++ b/nx/blocks/form/views/editor.js
@@ -40,8 +40,15 @@ class FormEditor extends LitElement {
   }
 
   handleChange({ target }) {
-    const { name } = target; // pointer
-    let value = target.type === 'checkbox' ? target.checked : target.value;
+    const { name } = target;
+    let value;
+    switch (target.type) {
+      case 'checkbox':
+        value = target.checked;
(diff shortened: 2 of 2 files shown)


nx/blocks/form/utils/generator.js
```diff
@@ -25,10 +25,8 @@ export default function generateEmptyObject(
     return generateEmptyObject(schema.oneOf[0], requiredFields, rootSchema);
   }
 
-  // If field has enum values, return the first one if it's required
-  if (schema.enum && schema.enum.length > 0 && requiredFields.size > 0) {
-    return schema.enum[0];
-  }
```
kozmaadrian: As agreed, we will not assign any default values other than those explicitly defined in the schema. The same approach applies below.

nx/blocks/form/utils/generator.js
```diff
@@ -25,10 +25,8 @@ export default function generateEmptyObject(
...
+  // Use schema default when present, otherwise undefined (no auto-picking enum[0] or 0)
+  const useDefault = (value) => (value !== undefined && value !== null ? value : undefined);
```
chrischrischris: nit: can use nullish coalescing
```suggestion
  const useDefault = (value) => value ?? undefined
```

hannessolo:
> Boolean values don't seem to work, the check box appears but doesn't add a value to the json output

Now fixed

> For some reason the UI isn't displaying the property names

I was missing the title property in the schema

> The navigation is completely empty

Same, title missing

PR #203, opened 2026-02-25
feat: [dasc] Add basic validation.

kozmaadrian:
Ticket: https://github.com/adobe/da-nx/issues/184

---

### Summary

The form was refactored so there's one place that holds all the data, the annotation tree was slimmed down, and validation plus default handling were added.

---

### Data and Structure Separation

All form values now live in a single JSON object. The annotation tree only describes layout and metadata. It doesn't hold values anymore.

One place was needed to look for data. So values were moved into the model and read by pointer instead of being stored in the annotation. That same structure makes validation (errors by field) and default value merging straightforward.

---

### Slim Annotation Tree

The annotation tree no longer carries raw schema. It only keeps what the UI needs: type, title, enum, required, pointers, and children.

Raw schema is messy and inconsistent. Components were full of schema-specific logic. The slim tree gives them a stable shape and keeps schema handling in one place.

---

### Validation

Form data is validated against the schema and required fields are checked so empty required ones are flagged. Each error is tied to a field so the editor can show it on the right input.

Validation is used to give feedback before save. Schema validation handles type and enum rules; a separate pass handles required empty fields. Using the same pointers as the model keeps errors aligned with the right fields. Saving is not blocked when validation fails; users can still save invalid content.

---

### Default Values

Schema defaults fill in fields when the user hasn't set anything. User input always overrides defaults. When merging, saved values are kept and defaults only fill gaps where there's no user value.

**Cases**

- **Empty form (new document):** Fields show schema defaults when they exist. Required fields get defaults. Optional ones use them if defined, otherwise stay empty.

- **User saves without edits:** The merged data, including defaults, is saved. Pruning only removes truly empty values. Defaults are treated as real values and get saved.

- **User reopens a saved form:** What was saved is shown. Omitted fields stay omitted. Schema defaults are not reapplied on load.

- **User edits a field that had a default:** Whatever the user sets replaces the default, including empty. That value is what gets saved and loaded. The schema default no longer applies for that field.

- **User clears an optional field (e.g. selects "None"):** The field is omitted from the data. On reload, schema defaults are not reapplied. The field stays empty. Defaults are only used when creating new structure (new form or new array item).

### DIFF
--- a/nx/blocks/loc/connectors/trados/auth.js
+++ b/nx/blocks/loc/connectors/trados/auth.js
@@ -1,0 +1,69 @@
+import { BASE_OPTS, corsFetch } from './utils.js';
+
+const TOKEN_BUFFER = 300000; // 5 min buffer before expiry
+
+// --- localStorage helpers ---
+
+function tokenKey(name, env) {
+  return `${name.toLowerCase()}.${env}.token`;
+}
+
+function getTokenDetails(name, env) {
+  const stored = localStorage.getItem(tokenKey(name, env));
+  if (stored) {
+    try {
+      return JSON.parse(stored);
+    } catch {
+      return {};
+    }
+  }
+  return {};
+}
+
+function setTokenDetails(name, env, accessToken, expires) {
+  localStorage.setItem(tokenKey(name, env), JSON.stringify({ accessToken, expires }));
+}
+
+// --- Exports ---
+
+// Returns cached token or fetches new one via Auth0 client credentials flow
+export async function getAccessToken(service) {
+  const {
+    name,
+    env = 'prod',
+    clientId,
+    clientSecret,
+    authEndpoint,
+    audience,
+  } = service;
+
+  // Return cached token if still valid
+  const { accessToken: cached, expires: cachedExpires } = getTokenDetails(name, env);
+  if (cached && cachedExpires > Date.now()) return cached;
+
+  const body = JSON.stringify({
+    client_id: clientId,
+    client_secret: clientSecret,
+    grant_type: 'client_credentials',
+    audience,
+  });
+
+  const opts = { ...BASE_OPTS, body };
+  const resp = await corsFetch(authEndpoint, opts);
+  if (!resp.ok) return null;
+
+  const json = await resp.json();
+  const { access_token: accessToken, expires_in: expiresIn } = json;
+  if (!accessToken) return null;
+
+  const expires = Date.now() + (expiresIn * 1000) - TOKEN_BUFFER;
+  setTokenDetails(name, env, accessToken, expires);
+
+  return accessToken;
+}
+
+// Ensures we have a valid token, returns true/false
+export default async function authReady(service) {
+  const accessToken = await getAccessToken(service);
+  return !!accessToken;
+}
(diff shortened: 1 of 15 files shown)


nx/blocks/form/views/editor.js
```diff
@@ -69,54 +90,69 @@ class FormEditor extends LitElement {
...
-    const opts = { detail: { pointer, itemsSchema }, bubbles: true, composed: true };
+    const { pointer, items } = parent;
+    const opts = { detail: { pointer, items }, bubbles: true, composed: true };
     this.dispatchEvent(new CustomEvent('add-item', opts));
   }
 
   renderCheckbox(item) {
-    const label = `${item.schema?.title ?? ''}${item.required ? ' *' : ''}`;
+    const error = this.getError(item.pointer);
+    const label = `${item.title ?? ''}${item.required ? ' *' : ''}`;
```
kozmaadrian: Inputs no longer own schema knowledge, they consume simple annotated structures instead, like `item.title`

nx/blocks/form/utils/utils.js
```diff
@@ -119,17 +193,15 @@ export async function loadHtml(details) {
...
  * @param {Object} propSchema - Schema for this property
  * @param {Object} fullSchema - Full schema for $ref resolution
- * @param {*} userData - User data at this level (may be undefined)
+ * @param {*} userData - User data at this level (for array length; values not stored)
  * @param {string} parentPointer - Parent pointer (e.g. "/data")
  * @param {boolean} required - Whether property is required
- * @returns {Object} { key, data, schema, pointer, required }
+ * @returns {Object} { key, pointer, title, type, enum?, items?, required, children? }
  */
 export function annotateFromSchema(key, propSchema, fullSchema, userData, parentPointer = '', required = false) {
```
kozmaadrian: The annotation mechanism has been refactored introducing the following changes:

---

Previously, the annotation mechanism produced an object that also stored data, effectively duplicating the data already present in the model JSON. In some cases, it could even contain more data, since the JSON loaded from the document never returns empty values (they are stripped out during persistence), while the annotated object could still add default values.

This created inconsistency and ambiguity around which structure should be considered the source of truth.

We need to clearly define a single source of truth for the data consumed by the UI when retrieving values. That same data source must also be used by other functionalities, such as validation. The annotated object was not suitable for validation purposes and could not be consumed consistently across the system.

The model JSON is now established as the single source of truth for all data.

The annotation object is used exclusively to define structure and UI related metadata, and no longer stores or duplicates data.

---

In addition, the annotation object previously exposed the full schema of a given property, and components such as the editor and navigation were responsible for handling all the different schema definition scenarios. Furthermore, adding all this detail to the annotated object caused it to grow significantly, including data that was not actually used.

With this change, the annotation object simply stores minimal, structured schema information. This provides a clean and predictable structure that underlying components can consume without needing to handle complex schema-specific cases.

---

Example schema and the resulted annotated object:

[annotated_object_PR_203.json](https://github.com/user-attachments/files/25569975/annotated_object_PR_203.json)
[schema_PR_203 schema.json](https://github.com/user-attachments/files/25569976/schema_PR_203.schema.json)

auniverseaway (approved):
Try to use domain-friendly URLs so we can test easier.

https://issue-184-b--da-nx--adobe.aem.live <-- this would be `issue/184_b`.

Issue #227, opened 2026-03-17
[dasc] Root array schema saved as object instead of array


kozmaadrian:
## Problem

The root array item is converted as an object instead of an array.

## Root cause

The JSON↔HTML conversion treats root arrays like objects: it uses numeric keys (`0`, `1`) as row labels in the HTML instead of the `@items` structure used for nested arrays. When parsed, each row becomes a key-value pair, producing an object instead of an array.

---

### Schema (root is array)

```json
{
  "type": "array",
  "title": "Color Families",
  "items": {
    "type": "object",
    "properties": {
      "name": { "type": "string" },
      "groups": { "type": "array", "items": { "type": "array", "items": { "type": "string" } } }
    }
  }
}
```

### JSON — correct vs incorrect

| | Correct | Incorrect |
|---|---|---|
| **data** | `[{ "name": "cf 1", ... }, { "name": "cf 2", ... }]` | `{ "0": { "name": "cf 1", ... }, "1": { "name": "cf 2", ... } }` |

### HTML — incorrect format (as saved)

Root block uses `0`, `1` as labels instead of `@items`:

```html
<div class="color-families">
  <div>
    <div><p>0</p></div>
    <div><p>self://#0-qbjyv0</p></div>
  </div>
  <div>
    <div><p>1</p></div>
    <div><p>self://#1-d3ias2</p></div>
  </div>
</div>
```

PR #228, opened 2026-03-17
fix: [dasc] Root array schema saved as object instead of array

kozmaadrian:
Ticket: https://github.com/adobe/da-nx/issues/227

## PR description

Root arrays are now stored and loaded as arrays. Includes a temporary migration for existing content.

**JSON → HTML:** Root arrays use the `@items` structure (one row, list of block refs), same as nested arrays. No more numeric keys as labels.

**HTML → JSON:** Root block = single class; item block = double class. Both share the schema name as the first class, so we use class count to tell them apart: the root container has one class (e.g. `color-families`), each array item has two (e.g. `color-families color-families-abc123`). Without this, item blocks could be misidentified as root.

**Migration:** Existing content was saved in the wrong format (object with numeric keys). Without conversion on load, that content would fail validation and break the form. The migration converts it to an array when loading from HTML so users can open, edit, and resave. Once resaved, the content is stored correctly. Remove the migration after customers have resaved their content.

### DIFF
--- a/nx/blocks/scheduler/scheduler.css
+++ b/nx/blocks/scheduler/scheduler.css
@@ -1,0 +1,158 @@
+:host {
+  display: block;
+  max-width: var(--grid-container-width);
+  margin: var(--spacing-800) auto;
+
+  h1 {
+    line-height: 1;
+    margin-bottom: 12px;
+  }
+
+  nx-path {
+    display: block;
+    margin-bottom: 16px;
+  }
+
+  p {
+    margin: 0;
+  }
+}
+
+.scheduler-header {
+  margin-bottom: 16px;
+}
+
+.nx-alert {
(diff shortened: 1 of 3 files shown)


nx/blocks/form/utils/value-resolver.js
```diff
@@ -28,7 +26,7 @@ export function generateValue(node, useSchemaDefaults = true) {
     case 'object':
       return generateObject(node, useSchemaDefaults, generateValue);
     case 'array':
-      return ARRAY_FALLBACK_VALUE;
+      return [];
```
kozmaadrian: Completely overlooked, this caused all cases using the fallback value to share the same empty array as their initial value.

mhaack (approved):

PR #293, opened 2026-03-30
Feat: Snapshot Enhancements

chrischrischris:
* Launch is now disabled by default.  To enable launch add `key: snapshot.launch, value: 'true'` to the DA site or org config
* New "Find Fragments" button will search through all urls and look for fragments which can then be added to the url list
* New per-url options: When a url is clicked multiple options are now displayed for that url: ** Open on aem.reviews
** Open on aem.live
** Edit in DA
** Edit Snapshot in DA (Only if snapshot exists)
** Sync Down (Only if snapshot.launch is enabled)
** Promote Up (Only if snapshot.launch is enabled and snapshot exits)

Test URLs:
- Before: https://da.live/apps/snapshots#/aem-sandbox/block-collection
- After: https://da.live/apps/snapshots?nx=snap-frag#/aem-sandbox/block-collection

### DIFF
--- a/nx/blocks/media-library/core/urls.js
+++ b/nx/blocks/media-library/core/urls.js
@@ -207,6 +207,40 @@ export function isPreviewPreferredForMediaUrl(mediaUrl) {
   return !!key && previewPreferredMediaKeys.has(key);
 }
 
+/** Normalize HLX hostnames to AEM delivery suffixes for comparison. */
+function hlxToAemHost(hostname) {
+  if (!hostname) return '';
+  return hostname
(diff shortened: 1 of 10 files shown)


chrischrischris:
@sukamat Right now just in the snapshot ui.  I'm not sure that snapshot usage is enough to warrant being part of the UI

auniverseaway:
@sukamat I think if someone needs this in the editor, they can make a Prepare menu plugin. If we get traction on this, we could bubble this up to DA. It gets a little messy: having to iterate through every snapshot to see if the page is in each, but it's not unreasonable.

PR #317, opened 2026-04-04
fix: simplied PDF view handling 

kmurugulla:
Removes PDF blob loading complexity (~200 lines) and renders PDFs like fragments (icon + colored
  background). Fixes localhost URLs in parsing by converting to sidekick config preview host and handles double-encoded
  filenames. Consolidates duplicate helpers and removes dead code.

Fix #<gh-issue-id>

Test URLs:
- After: https://da.live/apps/media-library?nx=mlpdfprevtab#/aemsites/vitamix


nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -736,15 +729,11 @@ class NxMediaInfo extends LitElement {
     if (this.media) {
       const label = getMediaName(this.media);
       if (label && label !== 'Unknown') {
-        displayName = label;
+        displayName = decodeDisplayName(label);
```
amol-anand: **Bug:** `getMediaName()` already calls `decodeDisplayName()` internally (`templates.js:62`). Wrapping the result again means up to 6 decode iterations total, which will over-decode legitimate percent-encoded names.

```suggestion
        displayName = label;
```

nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -71,6 +71,8 @@ class NxMediaInfo extends LitElement {
     _imageDimensions: { state: true },
     _comprehensiveMetadata: { state: true },
     _modalNotification: { state: true },
+    _pdfState: { state: true },
+    _pdfError: { state: true },
```
amol-anand, line 76: `_pdfBlobUrl` is checked in the render to decide whether to show the iframe, but it's not declared as reactive state. This works today only because `_pdfState` is set *after* `_pdfBlobUrl` in `loadPdfBlob` and triggers the re-render. Any future refactor changing that assignment order will silently break rendering.

```suggestion
    _pdfState: { state: true },
    _pdfError: { state: true },
    _pdfBlobUrl: { state: true },
```

nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -940,55 +929,47 @@ class NxMediaInfo extends LitElement {
...
-            class="pdf-preview"
-            @load=${this.handlePdfLoad}
-            @error=${this.handlePdfError}
-          >
-          </iframe>
-          <div class="document-placeholder">
-            <svg class="icon" viewBox="0 0 20 20">
+          <div class="pdf-loading-container" role="status" aria-live="polite">
+            <sl-spinner style="font-size: 3rem;" aria-label="Loading PDF"></sl-spinner>
+            <p class="pdf-error-message">Loading PDF preview...</p>
```
amol-anand: **i18n regression:** The old code used `t('UI_PDF_OPEN_IN_NEW_TAB')`, `t('UI_PDF_NETWORK')`, etc. All user-visible strings in the render and in `loadPdfBlob` are now hardcoded English:

- `"Loading PDF preview..."`
- `"Failed to load PDF"`
- `"Open PDF in new tab"`
- `"Unable to load PDF (${response.status})"` (line 418)
- `"Server did not return a PDF file"` (line 425)
- `"Failed to load PDF. It may be inaccessible."` (line 447)

These all need to go through `t()`.

nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -940,55 +929,47 @@ class NxMediaInfo extends LitElement {
...
+              Open PDF in new tab
+            </a>
           </div>
         `;
       }
 
-      const pdfError = this._pdfLoadErrors.get(fullUrl);
+      if (this._pdfBlobUrl) {
+        return html`
+          <iframe src="${this._pdfBlobUrl}" class="pdf-preview" title="PDF Preview" frameborder="0" aria-label="PDF document preview"></iframe>
```
amol-anand: `frameborder="0"` is deprecated in HTML5 and already handled by `border: none` in `.pdf-preview` CSS.

```suggestion
          <iframe src="${this._pdfBlobUrl}" class="pdf-preview" title="PDF Preview" aria-label="PDF document preview"></iframe>
```

nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -940,55 +929,47 @@ class NxMediaInfo extends LitElement {
...
-            src="${blobUrl}"
-            class="pdf-preview"
-            @load=${this.handlePdfLoad}
-            @error=${this.handlePdfError}
-          >
-          </iframe>
-          <div class="document-placeholder">
-            <svg class="icon" viewBox="0 0 20 20">
+          <div class="pdf-loading-container" role="status" aria-live="polite">
+            <sl-spinner style="font-size: 3rem;" aria-label="Loading PDF"></sl-spinner>
```
amol-anand: Inline `style="font-size: 3rem;"` appears on both spinner instances (here and line 967). Move to the stylesheet instead — e.g. a `.pdf-loading-container sl-spinner` rule.

nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -394,6 +383,72 @@ class NxMediaInfo extends LitElement {
...
+      const blob = await response.blob();
+
+      if (this._pdfCurrentUrl !== pdfUrl) {
+        return;
+      }
+
+      const blobUrl = URL.createObjectURL(blob);
+
+      this._pdfBlobUrl = blobUrl;
+      this._pdfBlobCache.set(pdfUrl, blobUrl);
```
amol-anand, line 452: Nit: `_pdfBlobCache` entries are never evicted until the component disconnects. If a user navigates through many PDFs in a session, blob URLs accumulate in memory. Consider revoking the previous entry when loading a new one (the cache-hit fast path is fine — it's the unbounded growth that's the concern).

nx/blocks/media-library/views/mediainfo/mediainfo.js
```diff
@@ -1,6 +1,6 @@
 import { html, LitElement } from 'da-lit';
 import getStyle from '../../../../utils/styles.js';
-import getSvg from '../../../../utils/svg.js';
+import loadSvgIcons from '../../../../utils/svg.js';
```
amol-anand, line 3: Nit: This rename is cosmetic and incomplete — `utils/svg.js` still exports `export default getSvg`. Either rename at the source (and all consumers) in a follow-up PR, or keep the original name here to avoid inconsistency.

```suggestion
import getSvg from '../../../../utils/svg.js';
```

nx/blocks/media-library/core/files.js
```diff
@@ -41,12 +41,32 @@ export function getFileName(url) {
     try {
       return decodeURIComponent(filename);
     } catch {
-      // If decoding fails (malformed encoding), return as-is
       return filename;
     }
   }
 }
 
+export function decodeDisplayName(name) {
```
amol-anand, line 49: This function has non-trivial loop logic handling edge cases (double-encoding, malformed URIs, falsy inputs). Worth adding unit tests — `decodeDisplayName` and `toExternalMediaEntry` currently have zero test coverage.

nx/blocks/media-library/indexing/parse.js
```diff
@@ -584,23 +584,35 @@ export async function buildUsageMap(pageEntries, org, repo, ref, onProgress, onB
...
+  try {
+    displayName = decodeURIComponent(info.name);
+  } catch {
+    // Keep original if decode fails
+  }
 
   return {
-    hash: normalizedUrl, // Use normalized URL as dedupe key for external media
-    url, // Keep original URL for display
+    hash: normalizedUrl,
```
amol-anand, line 608: The hash key now includes the canonicalized URL (`normalizeExternalVideoUrl(canonicalizeMediaUrl(url, …))`). If existing index entries were keyed by the non-canonical form, this will create duplicates rather than deduplicating. Does the indexer handle this gracefully, or does this need an index rebuild / version bump?

amol-anand (changes requested):
Nice cleanup — the state machine replacing three parallel Maps is a clear win, and the abort controller + stale-request guard is well done. See inline comments for specifics; the two blocking items are the double-decode bug and the i18n regression.

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Fixed. Removed the outer decodeDisplayName() call since getMediaName() already decodes internally.

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Fixed. Added 6 new message keys to core/messages.js (UI_PDF_LOADING, UI_PDF_FAILED, UI_PDF_HTTP_ERROR,UI_PDF_INVALID_TYPE, UI_PDF_INACCESSIBLE, UI_PDF_LOADING_ARIA) and updated all hardcoded strings to use t() function.

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Fixed. Added _pdfBlobUrl: { state: true } to static properties declaration.

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Removed frameborder="0" attribute. Border styling is handled by CSS (.pdf-preview has border: none).

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Moved font-size: 3rem to CSS (.pdf-loading-container sl-spinner selector) and removed inline styles.

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Implemented LRU eviction with MAX_PDF_CACHE_SIZE=5. When cache is full, oldest blob is revoked and
  removed before adding new entry. Cache is also fully cleared on modal close via _resetState().

nx/blocks/media-library/views/mediainfo/mediainfo.js
kmurugulla: Kept as loadSvgIcons. ESLint rule import/no-named-as-default prevents using getSvg as the import name since it's both a default and named export. The rename is necessary to pass lint checks.

nx/blocks/media-library/core/files.js
kmurugulla: Added comprehensive unit tests in test/media-library/core/files.test.js covering
single/double/triple encoding, malformed input, unicode, and edge cases.

nx/blocks/media-library/indexing/parse.js
kmurugulla: Verified. Deduplication works correctly with the org/repo parameters. 
Flow: toExternalMediaEntry() receives org/repo → passes to canonicalizeMediaUrl(url, org, repo) → normalizes localhost/internal URLs consistently → getDedupeKey() extracts pathname/filename for deduplication. 
  
 The org/repo params ensure URLs from the same site are canonicalized identically before deduplication, preventing false duplicates while catching true ones.

amol-anand (approved):

PR #318, opened 2026-04-05
fix:migrated from custom state management to Lit's reactive props

kmurugulla:
Test URLs:
- After: https://da.live/apps/media-library?nx=mlstate#/aemsites/vitamix


kmurugulla:
Used Slicc for analysis with the following prompt when i have Media Library opened with main and state migrate branch 

`I have two Media Library Tabs .

First one is with a branch where the code is depending on Lit State
Second with a custom state management
https://github.com/adobe/da-nx/pull/318 has the changes i made
While this reduces number of lines of code , I want to confirm whether this is the worth clean up

-Without causing any regression
-how many lines of code are we saving
-Is depending on Lit state a better dev or design principle
-Are the operations like Search , Index , Incremental index ( you have to refresh the page to kick in incremtal index) is any faster
-Is opening the modal , or navigation is any faster
-We have the state being pushed into url to create deep links . Are those operations any faster .

Please test and give me report as a table to make the decision .`

And the report shows that custom state is much better 

image

So closing this PR.

PR #327, opened 2026-04-08
Experience Workspace panels plumbing

hannessolo:
**Test URL**
https://da.live/canvas?nx=ew-layout&nxver=2

For pagespeed only
https://ew-layout--da-nx--adobe.aem.live/


nx2/blocks/canvas-actions/canvas-actions.js
```diff
@@ -0,0 +1,22 @@
+import { LitElement, html } from 'da-lit';
```
hannessolo, line 1: @auniverseaway Need your opinion here. We need to have a web component in the shell nav. This component is super specific to our `canvas` block though. Might be nicer to provide a way for the nav to load web components without those components being their own block? Or how would you solve this?

nx2/blocks/tool-panel/tool-panel.js
```diff
@@ -0,0 +1,3 @@
+export default function decorate() {
```
hannessolo, line 1: Looking for opinions if this is a good way to handle it.

I would see this as the entry point for the right-side panel, providing the header that lets us switch between the different tools this will render. The individual tools rendered inside the panel will live elsewhere though.

nx2/blocks/canvas/nx-canvas-header/nx-canvas-header.js
```diff
@@ -0,0 +1,82 @@
...
+
+import { loadStyle } from '../../../utils/utils.js';
+import { loadHrefSvg } from '../../../utils/svg.js';
+
+const style = await loadStyle(import.meta.url);
+
+const ICONS_BASE = new URL('../../../img/icons/', import.meta.url).href;
+
+const ICONS = {
+  undo: `${ICONS_BASE}S2_Icon_Undo_20_N.svg`,
```
hannessolo, line 9: Also need your opinion here @auniverseaway . We can't use the icon loading function that creates an svg in html, since browser security doesn't allow loading those from localhost when on da.live.

But this is also a bit of a mess, if every web component is responsible to load and cache it's own icons...

sharanyavinod:
@hannessolo I think we need to hide the left rail from the shell as well in the edit view - to be handled later?

hannessolo:
> to be handled later?

Yes, let's do that separately

nx2/blocks/canvas/canvas.js
```diff
@@ -0,0 +1,60 @@
...
+const FRAGMENTS = {
+  before: 'https://da.live/fragments/exp-workspace/chat',
+  after: 'https://da.live/fragments/exp-workspace/tool',
+};
+
+async function addPanelHeader(aside) {
+  const { default: createPanelHeader } = await import('./nx-panel-header/nx-panel-header.js');
+  aside.querySelector('.panel-body').prepend(await createPanelHeader({
+    position: aside.dataset.position,
+    onClose: () => hidePanel(aside),
```
sharanyavinod, line 16: I would also consider adding a panel attribute here, so we dont rely on position for toggling/persisting states but actual identities of the panels. This would also help in the future when we want to add multiple panels.

sharanyavinod (approved):

PR #334, opened 2026-04-09
fix: remove static values and cleanup styles

sharanyavinod:
- Cleans up leftover static values and fixes minor style issues on integration
- Introduce option to insert buttons in panel header by consumers

Preview at https://da.live/canvas?nx=fix-ew-chat-integration&nxver=2#/aem-sandbox/block-collection/ew-test.html

### DIFF
--- a/nx2/blocks/chat/chat.js
+++ b/nx2/blocks/chat/chat.js
@@ -6,20 +6,14 @@ import './welcome/welcome.js';
 import { loadChatIcons } from './utils.js';
 
 const styles = await loadStyle(import.meta.url);
+const icons = await loadChatIcons({ add: 'Add', clear: 'RemoveCircle', copy: 'Copy', send: 'ArrowUpSend', stop: 'Stop' });
 
-const ICONS = {
-  add: 'Add',
-  clear: 'RemoveCircle',
-  copy: 'Copy',
-  send: 'ArrowUpSend',
-  stop: 'Stop',
-};
+const icon = (name) => icons?.[name]?.cloneNode(true);
 
 class NxChat extends LitElement {
   static properties = {
     messages: { type: Array },
     thinking: { type: Boolean },
-    _icons: { state: true },
     connected: { type: Boolean },
   };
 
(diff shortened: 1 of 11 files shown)


hannessolo:
Screenshot 2026-04-09 at 17 20 12

Is it intentional that the icons are not square, and smaller than before?

nx2/blocks/chat/chat.js
```diff
@@ -34,13 +28,29 @@ class NxChat extends LitElement {
...
+    if (!this._panelSlot || this._clearBtn) return;
+    const btn = document.createElement('button');
+    btn.className = 'panel-header-action';
+    btn.setAttribute('aria-label', 'Clear chat');
+    btn.hidden = !this.messages?.length;
+    if (icons.clear) btn.append(icon('clear'));
+    btn.append(Object.assign(document.createElement('span'), { textContent: 'Clear' }));
+    btn.addEventListener('click', () => this.clear());
+    this._clearBtn = btn;
+    this._panelSlot.append(btn);
```
hannessolo, line 42: Hm, this seems a bit of an anti-pattern, are we sure we want to be modifying the dom imperatively if we otherwise render components declaratively?

hannessolo: Oh, I see. We're adding it to the header. That explains why but I still wonder if we can find a simpler way - maybe to at least get rid of the event?

sharanyavinod: Yes I'm not happy about it either. Every other approach I could think of have tradeoffs too. Would like your thoughts: 
1. Cleanest would be positioning via css - no rewiring, everything stays as before. BUT, we would need to explicitly position based on the size and position of the collapse button in each panel and seems too specific.
2. Make panel header creation configurable, but then we expose internals of each consumer and not ideal.
3. Else consumer owns panel header and panel passes in via slot or similar for the toggle button option - but then we have no assurance consumers actually implement it

sharanyavinod: Just to get rid of the event I can query the element directly, but have to switch things around a bit to make sure header is always loaded before content. Wdyt?

sharanyavinod:
> Is it intentional that the icons are not square, and smaller than before?

Smaller yes, square I was just fixing.

hannessolo (approved):

PR #339, opened 2026-04-10
feat: reusable menu and popover

sharanyavinod:
- Simple reusable popover container to be used by other blocks like menu, picker etc
- Menu block with a trigger slot for the button + support for divider and section title

Preview available at https://da.live/canvas?nx=ew-chat-menu&nxver=2#/aem-sandbox/block-collection/ew-test.html 
Look at the add option within the chat textarea

Screenshot 2026-04-10 at 14 50 24


nx2/blocks/shared/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+
+  get _popover() { return this.shadowRoot.querySelector('nx-popover'); }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.shadowRoot.adoptedStyleSheets = [styles];
+  }
+
+  firstUpdated() {
+    this._wireTrigger(this.shadowRoot.querySelector('slot[name="trigger"]'));
```
kozmaadrian, line 32: querySelector('slot[name="trigger"]') might return null

nx2/blocks/shared/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  async _loadIcons() {
+    const names = this.items?.map((i) => i.icon).filter(Boolean) ?? [];
+    if (!names.length) return;
+    this._icons = await loadSvgIcons(names);
+  }
+
+  _wireTrigger(slot) {
+    const [trigger] = slot.assignedElements();
```
kozmaadrian, line 46: slot might be null

nx2/blocks/shared/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+    if (!trigger || trigger === this._trigger) return;
+    this._trigger = trigger;
+    this._popover.anchor = trigger;
```
kozmaadrian, line 49: Could this._popover be null?

nx2/blocks/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+  _onMenuToggle(e) {
+    if (e.newState !== 'open') return;
+    this._active = undefined;
+    this._trigger?.toggleAttribute('data-active', true);
+    this._trigger?.setAttribute('aria-expanded', 'true');
+  }
+
+  show({ anchor, placement } = {}) {
+    this._active = undefined;
+    this._popover.show({
```
kozmaadrian: Same here, could this._popover be null? Maybe it worth to evaluate all cases when this._popover is used (e.g. this._popover.close() )

nx2/blocks/shared/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+        <button
+          role="menuitem"
+          data-id=${item.id}
+          class="menu-item ${item.id === this._active ? 'menu-item-active' : ''}"
+          type="button"
+          @click=${() => this._select(item)}
+          @mouseenter=${() => { this._active = item.id; }}
+          @focus=${() => { this._active = item.id; }}
+        >
+          ${item.icon && this._icons?.[item.icon] ? html`<span class="menu-item-icon">${this._icons[item.icon].cloneNode(true)}</span>` : nothing}
```
kozmaadrian, line 138: In case _loadIcons() returns with `if (!names.length) return;` then this._icons will not be initialised

nx2/blocks/shared/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+
+const ICONS_BASE = new URL('../../img/icons/', import.meta.url).href;
+const styles = await loadStyle(import.meta.url);
+
+// todo: remove once changes from ew are available with reusable utils
+export async function loadSvgIcons(names) {
+  const svgs = await Promise.all(
+    names.map((name) => loadHrefSvg(`${ICONS_BASE}S2_Icon_${name}_20_N.svg`)),
+  );
+  return Object.fromEntries(names.map((name, i) => [name, svgs[i]]));
```
kozmaadrian, line 14: loadSvgIcons always maps each name to svgs[i] even when loadHrefSvg fails

nx2/blocks/menu/menu.js
```diff
@@ -0,0 +1,157 @@
...
+
+    const curIdx = selectable.findIndex((i) => i.id === this._active);
+
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
+      this._active = selectable[(curIdx + 1) % selectable.length].id;
+      this.shadowRoot.querySelector(`[data-id="${this._active}"]`)?.focus();
+    } else if (e.key === 'ArrowUp') {
+      e.preventDefault();
+      this._active = selectable[(curIdx - 1 + selectable.length) % selectable.length].id;
```
kozmaadrian: With this._active unset selectable.findIndex((i) => i.id === this._active) returns -1, then the first ArrowUp selects index length - 2

nx2/blocks/shared/menu/menu.js
sharanyavinod: No, it runs after firstUpdated and render runs before that which always creates a slot unconditionally within the code

nx2/blocks/shared/menu/menu.js
sharanyavinod: No, its called from either firstUpdated or on slotchange, first of which as mentioned in the comment above and second is change on the slot itself - so neither will be null

nx2/blocks/shared/menu/menu.js
sharanyavinod: Also no, popover get does queryselector for nx-popover which is rendered unconditionally similar to the slot

hannessolo:
@auniverseaway I think these 2 components would be a good fit for a `components` directory instead of `blocks`, what do you think? As we discussed last week.

nx2/blocks/shared/menu/menu.js
sharanyavinod: If there are no icon names, what would need to be initialised anyway?

nx2/blocks/shared/menu/menu.js
sharanyavinod: Sure, but it would ba falsy value and we check against this when rendering the icon

nx2/blocks/shared/menu/menu.js
kozmaadrian: In case this._icons is null this call with throw ${this._icons[item.icon].cloneNode(true)}

nx2/blocks/menu/menu.js
sharanyavinod: Nice catch, updated

nx2/blocks/menu/menu.js
sharanyavinod: Unlikely, but indeed good to err on the side of caution since they are public methods. Updated.

nx2/blocks/shared/menu/menu.js
sharanyavinod: How? If this._icons is null, it fails the ternary check and renders nothing

PR #349, opened 2026-04-13
feat: Add basic breadcrumb.

kozmaadrian:
Implements a basic breadcrumb navigation.

Test URLs: https://da.live/browse?nx=ew-breadcrumb&nxver=2#/kozmaadrian/da-sc

Screenshot 2026-04-13 at 3 41 55 pm


nx2/blocks/breadcrumb/breadcrumb.js
```diff
@@ -0,0 +1,69 @@
...
+                    `}
+              </li>
+            `)}
+        </ol>
+      </nav>
+    `;
+  }
+}
+
+if (!customElements.get('nx-breadcrumb')) {
```
hannessolo: do we need the if statement? In the other blocks, we simply defined it. If the file is imported multiple times, this is only executed once anyway.

nx2/blocks/browse/browse.js
```diff
@@ -96,12 +106,16 @@ class NxBrowse extends LitElement {
     }
 
     if (this._items === undefined) {
-      return nothing;
+      return html`
```
hannessolo: we should find a unified way to handle UI for loading states. The DA/EDS pattern so far is to show nothing during loading.

I'm not sure that will be what we end up with but maybe until we decide on a proper consistent UX you could also just show nothing here.

hannessolo:
From a UI perspective, I think a) the font is too small and b) the chevrons should be aligned centered to the text or so the bottom of the icon is touching the baseline, now they're too high up.

nx2/blocks/browse/browse.js
kozmaadrian: For now I've reverted it to simply return "nothing"

kozmaadrian (commented):
I’ve applied the latest UX updates, including adjustments to font, spacing, and element positioning.

nx2/blocks/browse/browse.css
```diff
@@ -39,4 +39,35 @@
...
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    font-size: var(--s2-heading-size-m);
+    line-height: 26px;
+    font-weight: var(--s2-component-m-bold-font-weight);
+    color: var(--s2-gray-900);
+  }
+
+  nx-breadcrumb {
+    display: block;
```
sharanyavinod: Why is this required?

nx2/blocks/breadcrumb/breadcrumb.js
```diff
@@ -0,0 +1,62 @@
...
+
+class NxBreadcrumb extends LitElement {
+  static properties = {
+    pathSegments: { type: Array, attribute: false },
+    _chevronSvg: { state: true },
+  };
+
+  constructor() {
+    super();
+    this.pathSegments = undefined;
```
sharanyavinod: We want to avoid such constructors I believe

nx2/blocks/breadcrumb/utils.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+  const n = pathSegments.length;
+  return pathSegments.map((label, i) => {
+    if (i === n - 1) {
+      return { label };
+    }
+    if (i === 0 && n >= 2) {
+      return { label, href: `#/${pathSegments[0]}/${pathSegments[1]}` };
+    }
+    if (i === 0) {
```
sharanyavinod: Would this ever execute? If n is 0, it returns at L11. If i is 0 and n is 1 L15 passes, and if n is >=2 L19 passes. This seems to be dead code, can you check?

nx2/blocks/breadcrumb/breadcrumb.js
```diff
@@ -0,0 +1,62 @@
...
+
+    return html`
+      <nav class="breadcrumb" aria-label="Breadcrumb">
+        <ol>
+          ${crumbs.map((crumb, i) => html`
+              <li class="crumb">
+                ${i > 0 && this._chevronSvg
+        ? html`
+                      <span class="crumb-separator" aria-hidden="true">
+                        ${this._chevronSvg.cloneNode(true)}
```
sharanyavinod: Since the chevron is purely decorative, why not add via css instead? Can get rid of all the updates via js and async renders. Wdyt?

nx2/blocks/browse/browse.css
kozmaadrian: Removed.

nx2/blocks/breadcrumb/breadcrumb.js
kozmaadrian: Fixed.

nx2/blocks/breadcrumb/breadcrumb.js
kozmaadrian: Good idea.

anfibiacreativa (approved):
LGTM

sharanyavinod (approved):

PR #351, opened 2026-04-13
Ew quickedit

hannessolo:
https://da.live/canvas?nx=ew-quickedit&nxver=2#/hannessolo/author-kit/demo

https://ew-quickedit--da-nx--adobe.aem.page


nx2/blocks/canvas/canvas.js
```diff
@@ -52,11 +76,69 @@ export default async function decorate(block) {
   }
 
   const header = document.createElement('nx-canvas-header');
+  try {
```
hannessolo: TODO to myself - this file isn't cleaned up yet

nx2/blocks/canvas/nx-editor-doc/nx-editor-doc.js
```diff
@@ -0,0 +1,225 @@
+import { LitElement, html, nothing } from 'da-lit';
+import { loadStyle } from '../../../utils/utils.js';
+import { updateDocument, updateCursors } from '../nx-editor-wysiwyg/utils/document.js';
+import { getEditor } from '../nx-editor-wysiwyg/utils/state.js';
```
sharanyavinod: If these are shared across doc and wyswyg, maybe makes sense to have in a common location instead?
Ideally would be good if neither doc nor wyswyg files directly import from the other, and everything that is shared stays separate, wdyt?

nx2/blocks/canvas/canvas.js
```diff
@@ -1,14 +1,121 @@
...
+const WYSIWYG_PORT_READY_ATTR = 'data-nx-wysiwyg-port-ready';
+
+function applyCanvasEditorView(mountRoot, view) {
+  const v = normalizeCanvasEditorView(view);
+  const doc = mountRoot.querySelector('nx-editor-doc');
+  const frame = mountRoot.querySelector('nx-editor-wysiwyg');
+  if (doc) doc.hidden = v !== 'content';
+  if (frame) {
+    const portReady = frame.hasAttribute(WYSIWYG_PORT_READY_ATTR);
+    frame.hidden = v !== 'layout' || !portReady;
```
sharanyavinod: Instead of directly manipulating visibility here, could we instead fire an event saying what is currently active and let the consumers decide what to do with it? Keeps the scope of canvas limited to defining context, and leaving implementation detail to the panel itself.

nx2/blocks/canvas/canvas.js
```diff
@@ -1,14 +1,121 @@
...
+    || block.querySelector('.default-content')
+    || block;
+}
+
+function wireQuickEditPortToDoc(mountRoot, header) {
+  mountRoot.addEventListener('nx-wysiwyg-port-ready', (e) => {
+    const editor = mountRoot.querySelector('nx-editor-doc');
+    const port = e.detail?.port;
+    if (editor && port) {
+      editor.quickEditPort = port;
```
sharanyavinod: Same here, canvas should only ideally be in charge of getting the info and passing it on, how the doc uses it should be its internal detail

nx2/blocks/canvas/nx-editor-wysiwyg/nx-editor-wysiwyg.js
```diff
@@ -0,0 +1,185 @@
...
+  return {
+    config: {
+      mountpoint: `${getPreviewOrigin(org, repo)}/${org}/${repo}`,
+    },
+    location: { pathname },
+  };
+}
+
+function postQuickEditInitToIframe({ iframe, config, location, onReady }) {
+  const { port1, port2 } = new MessageChannel();
```
sharanyavinod: On each retry, we should probably also first close previously created ports that ended up unused.

nx2/blocks/canvas/nx-editor-wysiwyg/nx-editor-wysiwyg.js
```diff
@@ -0,0 +1,185 @@
...
+    const segments = path.split('/');
+    const pathWithoutOrgRepo = segments.slice(2).join('/');
+    const pathWithoutHtml = pathWithoutOrgRepo.replace(/\.html$/i, '');
+    const encodedPath = pathWithoutHtml.split('/').map(encodeURIComponent).join('/');
+    const base = `${getPreviewOrigin(org, repo)}/${encodedPath}?nx=exp-workspace&quick-edit=exp-workspace`;
+    return `${base}&controller=parent`;
+  }
+
+  _clearQuickEditRetry() {
+    if (this._quickEditInitRetryId) {
```
sharanyavinod, line 89: Would be nice to cleanup any open ports here as well

nx2/blocks/canvas/nx-editor-doc/nx-editor-doc.js
```diff
@@ -0,0 +1,225 @@
...
+    super.willUpdate(changed);
+    if (changed.has('ctx')) {
+      this._teardown();
+      this._error = undefined;
+    }
+  }
+
+  _clearControllerPort() {
+    if (this._controllerCtx?.port) {
+      this._controllerCtx.port.onmessage = null;
```
sharanyavinod: Similarly, close the port here as well on clear

nx2/blocks/canvas/canvas.js
sharanyavinod: canvas is still directly updating doc here

nx2/blocks/canvas/nx-editor-wysiwyg/nx-editor-wysiwyg.js
sharanyavinod: Similarly this doesnt seem to be updated either - while reset handles some cases with the newly introduced _disposeQuickEditLocalPort, would be good to have cleanup here as well since this is called directly from disconnectedcallback etc

nx2/blocks/canvas/canvas.js
hannessolo: Hm, I don't really see this as a problem. Isn't this just like a parent passing a prop to a child? What the child does with it is up to the child. Would you have the doc editor listen to the event directly instead?

sharanyavinod: > the child does with it is up to the child

Yes, so canvas shouldnt be concerned with setting it either. doc needs it, so it listens to it and handles it as required. Especially since in this case canvas isnt the parent owning the event, but rather wyswyg so canvas shouldn't take over this responsibility it has nothing to do with.
In other words, listen directly yes :)

sharanyavinod (approved):

PR #352, opened 2026-04-14
feat: introduce tools support

sharanyavinod:
### Summary

Introduces tool-call support in the Experience Workspace chat surface: the chat agent can request the editor to perform actions (open a page, insert content, etc.) and the user approves or rejects each call inline.

### Test URL

- Preview: https://da.live/canvas?nx=ew-tools&nxver=2#/hannessolo/author-kit/demo

### What's included

**Tool approval UX**
- Tool calls appear inline in the message stream with name, arguments and status.
- Calls that don't require user action (read-only / safe) auto-approve and render as a status row.
- Calls that require approval show a persistent popover above the chat input with Approve / Reject / Always-approve actions.
- Keyboard support: `↵` to approve focused, `Esc` to reject, `⌘↵` to always-approve for the conversation.

**State management**
- "Always approve" is conversation-scoped (persists per chat room, resets on new conversation).
- Message history is sanitised before send so tool payloads don't leak.
- Scroll-to-bottom on new tool events so approvals don't get hidden below the fold.

**Extensibility hooks**
- `tool-completed` event dispatched when a tool call resolves; current consumers use it for analytics, future ones will branch on payload (skills, follow-up tools, multi-step flows).
- Renderer registry in `chat/renderers.js` so new tool types can register their inline display without changing the chat controller.

### Documentation

Full agent stream contract and `da-agent` input-field expectations are documented in **`docs/chat-ui-component.md`**:
- Tool-call event shape (request, status updates, completion).
- Approval lifecycle (pending → approved/rejected → executed → completed).
- How to add a new tool renderer.


hannessolo:
The approval box is overlaying the input box, should it be above instead?

Screenshot 2026-04-15 at 11 44 12

hannessolo:
I also noticed when I refresh the page, the tool calls disappear.

sharanyavinod:
> The approval box is overlaying the input box, should it be above instead?
> 
It's overlaying by design - approval action is supposed to be blocking

sharanyavinod:
> I also noticed when I refresh the page, the tool calls disappear.

~~Yes, agent misbehaves when thats part of the conversation history. I can add it back purely for rendering if needed, but does it really add vaue from a conversation context perspective? Wdyt?
For me, it feels like that the cards are more for status while working and not really helpful as part of a conversation when visitng later, especially since the assistant summarizes what was done, what was the outcome etc.~~
On second thought, it would indeed be confusing for users to suddenly see things missing. I'll add it back, but atleast for things like read and so which didnt need any user input would be better shown simplified. I'll check with Kilian.

docs/chat-ui-component.md
```diff
@@ -30,11 +30,78 @@ The component manages its own controller internally. No external wiring needed.
...
+
+When the user clicks "Always approve", the tool name is added to an in-memory `Set` on the controller. Subsequent `tool-approval-request` events for that tool are auto-approved via `queueMicrotask`. The set is **conversation-scoped** — it resets only on `clear()`. There is no path-scoping: "always approve" applies to the tool by name regardless of what path the agent acts on, since the path is in the tool input rather than the tool name. There is no cross-session persistence.
+
+## Persistence
+
+Conversation history is persisted in IndexedDB, keyed by `org--site--userId`. This means:
+
+- History is **shared across all pages within a site** — navigating between paths does not start a new conversation.
+- History is **user-specific** — different IMS users on the same site have separate histories.
+- `clear()` deletes the stored history for the current room.
```
mhaack, line 112: This describes what you have today, but we maybe want to revise this in the future. The chat already as a concept of "room" today but we don't really use it for multiple rooms I think.

Especially for "shared across all pages within a site" we might need support for multiple rooms (in a different PR). Use Case: I do a review for accessibility or brand guiding on page A, do some changes and then I move into page B. At that point, at least today, the chat context still has all the information about page. Here we might either have a simple, clean chat or have a different room.

docs/chat-ui-component.md
```diff
@@ -30,11 +30,78 @@ The component manages its own controller internally. No external wiring needed.
...
+
+The approval popover accepts keyboard shortcuts: `Esc` = Reject, `↵` = Approve, `⌘↵` = Always approve.
+
+**If the agent team adds or renames event types, `processEvent` in `utils.js` must be updated to match.**
+
+### Approval summary rendering
+
+The UI picks one field from `input` to display as a human-readable summary beneath the tool name. Priority order:
+
+1. `humanReadableSummary` — preferred; plain-language description of what changed (used by `content_update`)
```
mhaack, line 91: Is this a proposal, or do we actually have this for one or all the tools?

hannessolo: We do have it implemented as a response from the agent, and were showing it in the UI at some point, although now the UI seems to be ignoring it.
Screenshot 2026-04-15 at 15 53 27

hannessolo: Sorry, wrong branch - it actually is shown by the UI on the exp-workspace branch:

Screenshot 2026-04-15 at 15 57 10

docs/chat-ui-component.md
sharanyavinod: Yes, I think that naturally comes in once we have the concept of sessions as in Kilian's mocks.  Do you think it makes sense to already include it now, with a default session?

mhaack: Lets keep it but mark is as planed/todo

docs/chat-ui-component.md
mhaack: On nice.

docs/chat-ui-component.md
sharanyavinod: @mhaack Updated - would be nice to also introduce a session title in the agent response when we introduce sessions. Have marked this as an open question.

docs/chat-ui-component.md
sharanyavinod: This is currently only available for content_update though

mhaack: Yes, lets keep in it the docs, we should have it for more of the tools

docs/chat-ui-component.md
```diff
@@ -30,11 +30,90 @@ The component manages its own controller internally. No external wiring needed.
...
+| Type | Fields | Description |
+|---|---|---|
+| `text-delta` | `delta` / `textDelta` / `text` | Incremental text chunk — appended to streaming buffer |
+| `text-end` | — | Flush streaming buffer as a committed assistant message |
+| `finish-message` / `finish` | — | Stream complete |
+| `error` | `errorText` / `error.message` | Agent error — thrown, caught by controller |
+
+### Tool events
+
+| Type | Aliases | Fields | Description |
```
anfibiacreativa: I think we could we normalize the naming conventions here to reduce ambiguity

Right now we have mixed pairs (input/args and output/result) plus alias event names, which makes the contract a little harder to reason about across producers/consumers. We could define one canonical shape (for example input and output) and explicitly document the others as backward-compat aliases (args, result) with guidance like “producers should emit canonical names; consumers should accept aliases.”

Not blocking, but I think this would make the API easier to implement

docs/chat-ui-component.md
```diff
@@ -30,11 +30,90 @@ The component manages its own controller internally. No external wiring needed.
...
+The controller consumes a server-sent event stream from `da-agent`. Each line is a JSON object with a `type` field. The UI depends on the following event types:
+
+### Text events
+
+| Type | Fields | Description |
+|---|---|---|
+| `text-delta` | `delta` / `textDelta` / `text` | Incremental text chunk — appended to streaming buffer |
+| `text-end` | — | Flush streaming buffer as a committed assistant message |
+| `finish-message` / `finish` | — | Stream complete |
+| `error` | `errorText` / `error.message` | Agent error — thrown, caught by controller |
```
anfibiacreativa: What is the precedence of errors between error events and `tool-result.output.error` ?

docs/chat-ui-component.md
```diff
@@ -30,11 +30,90 @@ The component manages its own controller internally. No external wiring needed.
...
+| Type | Aliases | Fields | Description |
+|---|---|---|---|
+| `tool-call` | `tool-input-available` | `toolCallId`, `toolName`, `input` / `args` | Agent invoked a tool |
+| `tool-approval-request` | — | `toolCallId`, `approvalId`, `toolName`, `input` | Tool requires user approval; `input` is the same args object from `tool-call` and is used to render the approval summary |
+| `tool-result` | `tool-output-available` | `toolCallId`, `toolName`, `output` / `result` | Tool completed; `output.error` signals failure |
+
+### Tool card states
+
+A tool card transitions through these states as events arrive:
+
```
anfibiacreativa, line 59: Could we document explicit ordering guarantees per toolCallId? The UI state assumes tool-call -> optional tool-approval-request -> tool-result, with tool-result as terminal. 

It would help to clarify whether this ordering is guaranteed, whether duplicates can occur, and how clients should handle out-of-order or missing events (especially with auto-approval and reconnect scenarios). I faced several race conditions when implementing the Skills lab feature...

sharanyavinod: Nice catch, will expand the docs a bit and recheck implementation.

sharanyavinod: @anfibiacreativa I've updated the doc, could you please check if the decisions are fair and aligned to the agent?

docs/chat-ui-component.md
sharanyavinod: Updated

docs/chat-ui-component.md
sharanyavinod: I've clarified the distinction

docs/chat-ui-component.md
anfibiacreativa: Look good now

nx2/blocks/chat/chat-controller.js
```diff
@@ -1,13 +1,17 @@
...
+import { AGENT_EVENT, ROLE, TOOL_STATE } from './constants.js';
 import { readStream } from './utils.js';
 import { loadMessages, saveMessages, clearMessages } from './persistence.js';
 
-const isLocal = new URLSearchParams(window.location.search).get('ref') === 'local';
-const AGENT_URL = isLocal ? 'http://localhost:5173/chat' : 'https://da-agent.adobeaem.workers.dev/chat';
+// ?ref=local routes to a local da-agent dev server (port 5173).
+const AGENT_URL = new URLSearchParams(window.location.search).get('ref') === 'local'
+  ? 'http://localhost:5173/chat'
+  : 'https://da-agent.adobeaem.workers.dev/chat';
```
anfibiacreativa, line 9: I had changed the port to 4200 because 5173 collided with Spellforge. I think in general I would choose other ports than Vite's default

nx2/blocks/chat/chat-controller.js
```diff
@@ -28,37 +32,58 @@ export default class ChatController {
...
+    // and cause "Tool result is missing". Complete approval sequences are kept so users
+    // see what the agent approved and did in prior conversations.
+    this._messages = cached.filter(
+      (m) => !(m.role === ROLE.ASSISTANT && Array.isArray(m.content)
+        && !m.virtual
+        && !m.content.some((p) => p.type === AGENT_EVENT.TOOL_APPROVAL_REQUEST)),
+    );
+    // Reconstruct tool cards from persisted approval messages so they render on reload.
+    this._toolCards = new Map();
+    for (const msg of this._messages) {
```
anfibiacreativa, line 46: For consistency and readability, maybe we want to stick to m, msg or message?

nx2/blocks/chat/chat-controller.js
```diff
@@ -28,37 +32,58 @@ export default class ChatController {
...
       await fetch(AGENT_URL, { method: 'HEAD', signal: AbortSignal.timeout(5000) });
       this._connected = true;
-      this._update();
     } catch {
       this._connected = false;
-      this._update();
       const delay = 1000 * 2 ** attempt;
-      if (delay >= 30000) return;
-      this._retryTimeout = setTimeout(() => this.connect(attempt + 1), delay);
+      if (delay < 30000) this._retryTimeout = setTimeout(() => this.connect(attempt + 1), delay);
```
anfibiacreativa, line 76: Is this retrying for ever? I can't see where we are maybe disconnecting or have max retries...

nx2/blocks/chat/renderers.js
```diff
@@ -42,27 +43,75 @@ function renderMessageContent(text) {
...
     </div>
   `;
 }
 
-export function renderMessage(msg, icons) {
-  if (msg.role === 'tool') return nothing;
-  const isAssistant = msg.role === 'assistant';
+function renderMessage(msg, icons, toolCards) {
+  if (msg.role === ROLE.TOOL) return nothing;
+  const isAssistant = msg.role === ROLE.ASSISTANT;
```
anfibiacreativa, line 95: I think in Anthropic's API, tools don't have a role. Could this be confusing at some point?

nx2/blocks/chat/chat-controller.js
sharanyavinod: Sure, I can update - just reused what was in the other branch at that point of time.

nx2/blocks/chat/chat-controller.js
sharanyavinod: Its checking up until delay hits 30s ie 5 times and then connected stays false. Wdyt - too short?

nx2/blocks/chat/renderers.js
sharanyavinod: Since our contract is with da-agent and it provides roles, I think thats the only thing the UI is concerned with. If at some point, the agent contract changes then we would need to revisit the implementation anyway, but until then I think this is setting the right expectations.
I had already noted this in constants file as an opportunity to have such items moved to a shared location and is a perfect example of how TS would help as we were discussing in yesterday's meeting  :).

nx2/blocks/chat/chat-controller.js
sharanyavinod: Done

nx2/blocks/chat/chat-controller.js
sharanyavinod: Done

hannessolo (approved):
From my side it looks good, the issues I had are fixed.

nx2/blocks/chat/chat-controller.js
anfibiacreativa: No, it's good. I had missed the disconnected set to false at the get go.

anfibiacreativa (approved):
LGTM!

PR #357, opened 2026-04-15
feat: reusable picker

sharanyavinod:
New reusable picker -> details in docs.

Preview available at https://da.live/canvas?nx=nx-picker&nxver=2#/hannessolo/author-kit/demo at the top of the chat ui.

Notes: 
- Shared util for keyboard navigation extracted to be reused by picker and menu. 
- Dummy code in chat.js enabled only for preview -> to be removed prior to merge.
- Checkmark for picker to be adapted to use svg icon once tools PR is merged

PR #360, opened 2026-04-16
Experience Workspace Slash Menu

hannessolo:
Screenshot 2026-04-16 at 12 29 12
Screenshot 2026-04-16 at 12 29 26

https://da.live/canvas?nx=ew-tools-v2&nxver=2#/hannessolo/author-kit/demo
https://ew-tools--da-nx--adobe.aem.page


nx2/blocks/shared/menu/menu.js
```diff
@@ -99,23 +103,36 @@ class NxMenu extends LitElement {
     this.dispatchEvent(new CustomEvent('select', { detail: { id: item.id }, bubbles: true, composed: true }));
   }
 
-  _onKeydown(e) {
+  handleKey(key) {
```
hannessolo, line 109: @sharanyavinod I actually think your function in https://github.com/adobe/da-nx/pull/357/changes#diff-8d969e1ebc4720381bc0e2b56d2753e2e24dd999089e2cb2a72d3ed5fc314548R1 is much cleaner than this, so my suggestino would be that we leave `handleKey` in place, but that `handleKey` internally calls that new function.

nx2/blocks/canvas/nx-editor-doc/slash-menu/slash-menu-handlers.js
```diff
@@ -0,0 +1,126 @@
...
+  wrapIn,
+  wrapInList,
+} from 'da-y-wrapper';
+
+/** From da-live `blocks/edit/prose/table.js` — empty table with heading row + 2×2 body. */
+function getTableHeading(schema) {
+  // eslint-disable-next-line camelcase
+  const { paragraph, table_row, table_cell } = schema.nodes;
+  const para = paragraph.create(null, schema.text('columns'));
+  // eslint-disable-next-line camelcase
```
sharanyavinod, line 15: Can we clean up as many of these as possible?

nx2/blocks/canvas/nx-editor-doc/slash-menu/slash-menu.js
```diff
@@ -0,0 +1,176 @@
...
+
+function destroySlashUi(ctxRef) {
+  ctxRef.hintEl?.remove();
+  const { ctx } = ctxRef;
+  if (!ctx) return;
+  ctx.menu.close();
+  ctx.scrollEl?.removeEventListener('scroll', ctx.onScroll);
+  ctx.anchor.remove();
+  ctx.menu.remove();
+  // eslint-disable-next-line no-param-reassign
```
sharanyavinod: The concern here is legitimate - we are clearing the passed object reference, but the original ctxRef in the closure isn't impacted and it would still contain hintEl and ctx. We probably need to clear those explicitly instead and then the eslint error is also gone since we mutate properties and not the object.

nx2/blocks/canvas/nx-editor-doc/slash-menu/slash-menu.js
```diff
@@ -0,0 +1,176 @@
...
+  // Anchor at paragraph start (the `/`) so the menu does not jump with the cursor while filtering
+  return { query, anchorPos: paraStart };
+}
+
+function shouldShowSlashHint(state) {
+  const { $from } = state.selection;
+  if (!inTopLevelParagraph($from)) return false;
+  if ($from.parentOffset !== 0) return false;
+  if ($from.parent.content.size > 0) return false;
+  if (getSlashContext(state)) return false;
```
sharanyavinod: Might be nice to chain them

sharanyavinod (approved):
Overall looks good, just a couple small comments.

nx2/blocks/canvas/nx-editor-doc/slash-menu/slash-menu-handlers.js
hannessolo: These are defined by the prosemirror schema, we can't change them

PR #368, opened 2026-04-17
feat: Enhance browse UI by adding extra columns

kozmaadrian:
Extend the browse UI by introducing additional columns

Test URL: https://da.live/browse?nx=ew-browsecolumns&nxver=2#/kozmaadrian/da-sc

columns


nx2/blocks/browse/browse-api.js
```diff
@@ -26,3 +27,110 @@ export async function listFolder(fullpath) {
...
+    .filter((i) => i >= 0);
+
+  for (let i = 0; i < indices.length; i += concurrency) {
+    const batch = indices.slice(i, i + concurrency);
+    await Promise.all(
+      batch.map(async (rowIndex) => {
+        try {
+          const item = out[rowIndex];
+          const path = itemToRepositoryPath(item, folderFullpath);
+          const json = await fetchResourceStatus(path, statusOrigin);
```
mhaack: Fetching individual page in parallel can lead into throttling on HLX, what if we have 1000 entries?

We should look into bulk status API, that is more efficient here. For a small number the bulk status is sync, only for larger numbers we need to handle the job status.

kozmaadrian: Yes, that would need to be handled in a separate PR.

nx2/blocks/browse/browse.js
```diff
@@ -56,23 +137,42 @@ class NxBrowse extends LitElement {
...
 
-  _onBrowseOpenFolder(event) {
-    const { pathKey } = event.detail;
-    if (!pathKey) {
+  _onBrowseActivate(event) {
+    const { pathKey, item } = event.detail || {};
+    if (entryTypeFromExtension(item.ext) === RESOURCE_TYPE.document) {
+      const url = new URL(window.location.href);
+      url.pathname = '/canvas';
+      url.hash = `#/${item.path.slice(1, -(item.ext.length + 1))}`;
```
kozmaadrian, line 86: When opening a document using /canvas, should the file extension be included in the path?

nx2/blocks/browse/browse.js
```diff
@@ -56,23 +137,42 @@ class NxBrowse extends LitElement {
...
+      window.location.assign(url.href);
+      return;
+    }
+    if (entryTypeFromExtension(item.ext) === RESOURCE_TYPE.sheet) {
+      const url = new URL(window.location.href);
+      url.pathname = '/sheet';
+      url.search = '';
```
kozmaadrian, line 93: The sheet is opened in the sheet editor using the default da-nx version.

nx2/blocks/browse/browse.js
kozmaadrian: Clarified, the URL should not include the file extension.

hannessolo (changes requested):
The way I understood the discussion, I thought we don't want to have the last previewed and last published displayed? 

1. I don't think this will be approved to merge into da if we have it
 2. We should check with the upcoming customer VIP projects whether this is something they need, and if yes we can start thinking about the bulk api to get the status.

kozmaadrian:
@hannessolo I've removed the columns, see the screenshot and the preview url.

nx2/blocks/browse/browse.js
```diff
@@ -1,12 +1,72 @@
...
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+  min-height: 0;
+}
+`.trim();
```
sharanyavinod: Why do we add styles from the js? If we do need to override certain things why not still keep styles in the stylesheet and include similar to  https://github.com/adobe/da-nx/blob/ew/nx2/blocks/canvas/canvas.js#L7-L8

nx2/blocks/browse/browse.js
```diff
@@ -1,12 +1,72 @@
...
+};
+
+const revertStyleOverride = () => {
+  if (!styleOverrideSheet) {
+    return;
+  }
+  document.adoptedStyleSheets = document.adoptedStyleSheets.filter(
+    (sheet) => sheet !== styleOverrideSheet,
+  );
+};
```
sharanyavinod: Adding at the module level would make these unnecessary too

nx2/blocks/browse/browse.js
kozmaadrian: Moved it to overriedes.css

nx2/blocks/browse/browse-api.js
kozmaadrian: In the end, we decided to remove the columns that call the status API for each row.

sharanyavinod (approved):

hannessolo (approved):

PR #375, opened 2026-04-21
feat: chat prompt picker

sharanyavinod:
Includes:

- Prompt picker which opens on "Load More" on the chat welcome screen if prompts are available AND on chat input add menu -> prompt
- Combobox includes options to filter by category and search by title/description

Preview available at https://da.live/canvas?nx=ew-prompts&nxver=2#/aem-sandbox/block-collection/ew-test

Screenshot 2026-04-21 at 11 45 05


hannessolo:
* Should we have an option to add the prompt to the chat instead of firing it directly?
* Should typing / in the chat box also open this prompt menu?

sharanyavinod:
> * Should we have an option to add the prompt to the chat instead of firing it directly?

Indeed would be nice. I'll check with Kilian as a followup, since if we add it would need to add on the welcome screen prompts as well.

> * Should typing / in the chat box also open this prompt menu?

As per both the mocks and the existing implementation on exp-workspace branch, slash menu shows the available DA tools and skills. The menu we have currently also cleanly separates prompts vs slash menu. Wdyt? Confusing?

Screenshot 2026-04-22 at 08 29 59

anfibiacreativa:
> * Should we have an option to add the prompt to the chat instead of firing it directly?
> * Should typing / in the chat box also open this prompt menu?

Actually this was the behavior from the prompt management and quick skills editor. That allows users to edit the prompts when needed.

nx2/blocks/chat/prompts/prompts.js
```diff
@@ -0,0 +1,123 @@
...
+  focus() {
+    this.shadowRoot.querySelector('.prompts-search-input')?.focus();
+  }
+
+  render() {
+    const total = this.prompts?.length ?? 0;
+    const filtered = this._filtered;
+    const placeholder = this._category === ALL_CATEGORY
+      ? `Search all ${total} prompts`
+      : `Search in ${this._category}`;
```
anfibiacreativa, line 84: How are we translating the UI? Or we don't?

nx2/blocks/chat/chat.js
```diff
@@ -2,27 +2,37 @@ import { LitElement, html, nothing } from 'da-lit';
...
 import { loadChatIcons } from './utils.js';
-import { ADD_MENU_ITEMS, CHAT_ICONS, ROLE, TOOL_STATE } from './constants.js';
+import { loadPrompts } from './api.js';
+import { ADD_MENU_ITEMS, CHAT_ICONS, MENU_OPTIONS, ROLE, TOOL_STATE } from './constants.js';
 
 const styles = await loadStyle(import.meta.url);
 const icons = await loadChatIcons(CHAT_ICONS);
 
 const icon = (name) => icons?.[name]?.cloneNode(true);
+const PROMPTS_GAP = 8;
```
anfibiacreativa: Should we prefix UI vars with UI_ so they're easier to find in code?

nx2/blocks/chat/chat.js
```diff
@@ -113,7 +129,7 @@ class NxChat extends LitElement {
 
   updated(changed) {
     if (changed.has('messages')) {
-      const log = this.shadowRoot.querySelector('.chat-messages-container');
+      const log = this.shadowRoot.querySelector('.chat-scroll-container');
```
anfibiacreativa, line 132: Should we move selectors we're querying to constants to prevent regressions?

hannessolo:
> Wdyt? Confusing?

I don't think it's confusing (although I'm not sure I understand what selecting a tool from the / menu will do), but I think it's too many clicks to run a prompt - click +, click prompts, select prompt

Vs if we integrated it with the / menu (slash would be combo of commands and skills), it's just type / and select prompt.

nx2/blocks/chat/prompts/prompts.js
sharanyavinod: There is no translation across DA afaik

nx2/blocks/chat/chat.js
sharanyavinod: This is an outlier so far, but I dont find any harm in doing so anyway. Will update.

nx2/blocks/chat/chat.js
sharanyavinod: Indeed, this is a recurring pattern. Will create a followup PR to collate all such usages across chat.

sharanyavinod:
> > Wdyt? Confusing?
> 
> I don't think it's confusing (although I'm not sure I understand what selecting a tool from the / menu will do), but I think it's too many clicks to run a prompt - click +, click prompts, select prompt
> 
> Vs if we integrated it with the / menu (slash would be combo of commands and skills), it's just type / and select prompt.

Regarding clicks, true I agree. But then the slash menu becomes a pot of all different things ie prompts,commands,skills  so that would need to be much cleaner and clearer than what we have atm. I would say we keep it as is for now, and follow up from the ux side whether these should also be present within the / menu. Adding it should be straightforward if we decide to do so anyway.

anfibiacreativa:
@sharanyavinod Do you want to start writing e2e tests with this feature?

sharanyavinod:
> @sharanyavinod Do you want to start writing e2e tests with this feature?

Do we start already, or post move to da live? 

cc @hannessolo

sharanyavinod:
I was also thinking - instead of a magical / which does everything, why not something like slack ie type/ shows the different options like /prompt /skill etc and selecting that opens the follow up menu with filter search etc like we have prompts today. Just an easier way of triggering prompts and others aligned to existing standards, and already includes control if any of these lists grows too much without overwhelming the end user. Thoughts?

Of course, nothing to be considered within the scope of this PR but more for when we implement slash menu.

hannessolo:
> Do we start already, or post move to da live?

I don't mind. Before we merge to da-live the tests are quite decoupled anyway.

mhaack:
In the Screenshot it says "DA Commands" this should be "Experience Workspace Commands" or if too long maybe just "Commands".

hannessolo:
> In the Screenshot it says "DA Commands" this should be "Experience Workspace Commands" or if too long maybe just "Commands".

I don't think the slash menu is implemented yet, that screenshot is from figma. We discussed in the daily today that we hold off the slash menu for now until we clarify where which types of items should show up (and whether we even need skills vs commands).

sharanyavinod:
@anfibiacreativa I discussed with @hannessolo as well, and I feel it might be too complicated to already start with tests while our changes are still across repos/branches. Would prefer to hold on adding them, until after we move to da-live.
For now, I would keep tests out of scope of this PR.

hannessolo (approved):

Issue #256, opened 2026-03-27
[EW Editor] Unified Toolbar


mhaack:
- Common text edit toolbar across all views (text formatting, link, assets, icons, placeholders)
- Toolbar smart? Different for images, blocks, RTE, links etc.
- Add support for new tools menu from doc based editor in canvas toolbar
- Tooltips on icon-only options

PR #380, opened 2026-04-21
Ew toolbar

hannessolo:
#256 

https://ew-toolbar--da-nx--adobe.aem.page

https://da.live/canvas?nx=ew-toolbar&quick-edit=ew-toolbar&nxver=2#/exp-workspace/frescopa/index


nx2/blocks/canvas/editor-utils/commands.js
```diff
@@ -0,0 +1,284 @@
...
+    if (node.isTextblock) {
+      visit(node, pos);
+      return false;
+    }
+    return true;
+  });
+}
+
+function everyTextblockInSelection(state, pred) {
+  let seen = false;
```
anfibiacreativa: Do we want to establish a convention for booleans naming like

isVisible
isOk

etc?

cc: @sharanyavinod

nx2/blocks/canvas/nx-editor-doc/slash-menu/slash-menu-handlers.js
```diff
@@ -110,17 +96,15 @@ function openLibraryPanel() {
...
   'insert-block': insertEmptyTable,
-  'heading-1': (state, dispatch) => setHeading(state, dispatch, 1),
-  'heading-2': (state, dispatch) => setHeading(state, dispatch, 2),
-  'heading-3': (state, dispatch) => setHeading(state, dispatch, 3),
-  blockquote: wrapInBlockquote,
-  'code-block': wrapInCodeBlock,
-  'bullet-list': (state, dispatch) => wrapInList(state.schema.nodes.bullet_list)(state, dispatch),
-  'numbered-list': (state, dispatch) => wrapInList(state.schema.nodes.ordered_list)(state, dispatch),
+  'heading-1': (s, d) => applyHeadingLevel(s, d, 1),
+  'heading-2': (s, d) => applyHeadingLevel(s, d, 2),
```
anfibiacreativa: Curious about the change from set to apply. Is it because it's not strictly a setter with a getter?

nx2/blocks/canvas/nx-editor-doc/nx-editor-doc.js
```diff
@@ -65,16 +65,17 @@ export class NxEditorDoc extends LitElement {
...
       port: this.quickEditPort,
+      iframe: this._wysiwygIframe,
+      suppressRerender: false,
       owner: org,
       repo,
-      pathname: controllerPathnameFromEditorCtx(this.ctx),
-      getToken,
-    });
+      path: controllerPathnameFromEditorCtx(this.ctx),
+      getToken: createQuickEditGetToken(),
```
anfibiacreativa: Couldn't this be just a generic utility getToken function?

nx2/blocks/canvas/editor-utils/commands.js
hannessolo: For locally scoped variables IMO we don't need this. But happy to be convinced otherwise. It would probably trigger renaming a lot of things though.

nx2/blocks/canvas/nx-editor-doc/slash-menu/slash-menu-handlers.js
hannessolo: It's just because I wanted a unified way of handling this between the slash menu and the toolbar. I think "apply" is more what we're doing than "set".

In the previous iteration of the code, it wasn't even consistently "set", I don't think `setBulletList` makes sense as a name

nx2/blocks/canvas/nx-editor-doc/nx-editor-doc.js
hannessolo: Yes, let me fix that. Please do call out more pointless functions like that. I had cursor refactor but it created a lot of wrapper functions that don't add value, I should have used a better model...

nx2/blocks/canvas/nx-editor-wysiwyg/utils/handlers.js
```diff
@@ -105,8 +109,39 @@ export function handleSelectionChange({ anchor, head }, ctx) {
...
+    return true;
   } catch (e) {
     // eslint-disable-next-line no-console
     console.error('[quick-edit-controller] handleSelectionChange failed', e?.message);
+    return false;
+  }
+}
+
+function getWysiwygIframe() {
+  return document.querySelector('nx-editor-wysiwyg')?.shadowRoot?.querySelector('iframe');
```
sharanyavinod: Doesnt ctx already have the iframe via wireQuickEditControllerPort? Couldnt we use that instead of querying again?

nx2/blocks/canvas/nx-editor-wysiwyg/utils/handlers.js
```diff
@@ -105,8 +109,39 @@ export function handleSelectionChange({ anchor, head }, ctx) {
...
+
+function positionSelectionToolbarFromIframe(data, ctx) {
+  const { view } = ctx;
+  const { anchorX, anchorY } = data;
+  const iframe = getWysiwygIframe();
+  if (!iframe) return;
+
+  const iframeRect = iframe.getBoundingClientRect();
+  const x = iframeRect.left + anchorX;
+  const y = iframeRect.top + anchorY - 64;
```
sharanyavinod: Since 64 is a repeated value for offset, maybe makes sense to move it to a constant?

nx/public/plugins/quick-edit/src/prose.js
```diff
@@ -160,6 +163,26 @@ function createEditor(cursorOffset, state, ctx) {
   setupImageDropListeners(ctx, editorParent);
   setRemoteCursors();
 
+  let scrollRaf = null;
+  editorParent.ownerDocument.defaultView.addEventListener('scroll', () => {
```
sharanyavinod: No cleanup for the listener?

nx2/blocks/canvas/editor-utils/selection-toolbar.js
```diff
@@ -0,0 +1,62 @@
...
+  if (toolbar) return toolbar;
+  componentLoaded ??= import('./nx-selection-toolbar.js');
+  toolbar = document.createElement('nx-selection-toolbar');
+  document.body.append(toolbar);
+  return toolbar;
+}
+
+export function getSelectionToolbar() {
+  return ensureToolbar();
+}
```
sharanyavinod: Why not use ensureToolbar directly?

nx2/blocks/canvas/editor-utils/commands.js
```diff
@@ -0,0 +1,284 @@
...
+
+export const EDITOR_TEXT_FORMAT_ITEMS = [
+  { id: 'heading-1', label: 'Heading 1', icon: 'Heading1' },
+  { id: 'heading-2', label: 'Heading 2', icon: 'Heading2' },
+  { id: 'heading-3', label: 'Heading 3', icon: 'Heading3' },
+  { id: 'blockquote', label: 'Blockquote', icon: 'BlockQuote' },
+  { id: 'code-block', label: 'Code block', icon: 'BlockCode' },
+  { id: 'bullet-list', label: 'Bullet list', icon: 'ListBulleted' },
+  { id: 'numbered-list', label: 'Numbered list', icon: 'ListNumbered' },
+];
```
sharanyavinod: Instead of having multiple array/map, could we have one list which collates items with their labels, icons etc. Would make it easier to maintain in the future if we have a single source of truth of supported commands.
Also, there are inconsistencies like `code-block`here and `code_block` elsewhere which would be good to avoid.
Also, imo this should be part of a separate constants/config file and not interspersed within the logic.

nx/public/plugins/quick-edit/src/prose.js
```diff
@@ -160,6 +163,26 @@ function createEditor(cursorOffset, state, ctx) {
...
+    if (scrollRaf) return;
+    scrollRaf = requestAnimationFrame(() => {
+      scrollRaf = null;
+      const { selection } = editorView.state;
+      if (selection.anchor === selection.head) return;
+      const offset = parseInt(editorParent.getAttribute('data-prose-index'), 10);
```
sharanyavinod: Might be worth validating offset value before proceeding here

sharanyavinod (changes requested):
We need a single source of truth for command config. The same information (id, schema name, label, icon etc) is currently spread across the code. Adding or renaming a command requires changes in multiple locations, and there's no overarching view of what's supported.

nx2/blocks/canvas/editor-utils/commands.js
anfibiacreativa: we don't need it locally, but it's a good way to consolidate and make code faster to understand.
We can start enforcing for other cases, though, and leave a big refactor as a TODO

hannessolo: I'm not convinced. I think "seen" and "ok" are already pretty obviously booleans (eg. `response.ok` is a really well known property in the fetch api). And anyway it would have to be `hasSeen` , not `isSeen`, in this case, because the function has seen it, not the function is being seen.

nx2/blocks/canvas/nx-editor-doc/nx-editor-doc.js
anfibiacreativa: I am somewhat convinced that we could make other functions in that same file that could be generic with more config tradeoff. But this one stood out.

nx2/blocks/canvas/nx-editor-wysiwyg/utils/handlers.js
anfibiacreativa: I am a fan of avoiding magic numbers at all costs

nx2/blocks/canvas/editor-utils/commands.js
anfibiacreativa: I also would propose that, if we really need to maintain this in code, we find a way a to put all these outside of the component program file, even if it stays in the same folder.

nx2/blocks/canvas/editor-utils/commands.js
anfibiacreativa: oh I was just giving examples, this is why I mentioned isVisible, not isSeen. But I get your point. Ok, is ok (pun intended) because it maps directly to a response.ok for a 200/299 range. Not sure seen is very intuitive I would call it hasTextBlocks or isNotEmpty. 
This is not a blocker to merge the PR, but it would be great to find our standards to work faster and have a style reference.

nx/public/plugins/quick-edit/src/prose.js
hannessolo: You're right, I somehow thought this was only happening once on initialization (in which case no teardown would be ok-ish) but it's actually being re-attached each time you edit something, for each editable item, not great. Let me improve that.

nx/public/plugins/quick-edit/src/prose.js
hannessolo: Hm, all of quick-edit relies on these values being correct (this is how we map content back from the iframe to the parent doc), so everything will break if they're wrong anyway - and there isn't really a way to recover from it.

nx/public/plugins/quick-edit/src/prose.js
hannessolo: Changed to only initialize once globally. We don't need a teardown for that case.

sharanyavinod (approved):

sharanyavinod:
FYI: I see you've been adding icons into https://github.com/adobe/da-nx/tree/ew/nx2/blocks/img/icons whereas we've been adding into https://github.com/adobe/da-nx/tree/ew/nx2/img/icons. Note for when you're moving things into da-live.

PR #385, opened 2026-04-23
Exp Workspace - Editor feature parity

hannessolo:
https://ew-plugins--da-nx--adobe.aem.page

https://da.live/canvas?nx=ew-plugins&quick-edit=ew-plugins&nxver=2#/exp-workspace/frescopa/index

Adds
- All prose mirror plugins as DA has
- Preview/publish button


sharanyavinod:
The coloring/sizing of the publish button seems really out of place with the rest of the page now
Screenshot 2026-04-23 at 15 11 52
Screenshot 2026-04-23 at 15 12 15

hannessolo:
> Agreed but I don't see any publish button in the UX designs so I just had to do something... better ideas?

Sizing -> update to match the profile button
Color -> I think maybe ok as is for now since it needs more attention, can check with Kilian once he's back

nx2/blocks/canvas/nx-editor-doc/prose.js
```diff
@@ -47,6 +63,24 @@ function addSyncedListener(wsProvider, canWrite, setEditable) {
   wsProvider.on('synced', handleSynced);
 }
 
+function handleUndo(state) {
+  const mgr = yUndoPluginKey.getState(state)?.undoManager;
+  if (mgr?.undoStack?.length > 0) {
+    mgr.undo();
+    return true;
+  }
+  return yUndo(state) || false;
```
sharanyavinod: Shouldn't yUndo be called with both state and dispatch here as well similar to in `undo()`?

nx2/blocks/canvas-actions/canvas-actions.js
```diff
@@ -1,19 +1,133 @@
...
     return html`
       <div class="canvas-actions">
-        <button>Publish</button>
+        <div class="right">
+          <div class="preview-row">
+            <button
+              type="button"
+              class="preview-dropdown-btn"
+              aria-label="Preview and publish"
+              aria-haspopup="dialog"
```
sharanyavinod: I think it should be menu and not dialog here since your popover content has role menu

nx2/blocks/canvas-actions/canvas-actions.js
```diff
@@ -1,19 +1,133 @@
...
+    if (!result.ok) {
+      this._error = formatAemPreviewPublishError(result.error);
+      this._busy = false;
+      this.requestUpdate();
+      return;
+    }
+
+    window.open(result.url, result.url);
+
+    this._busy = false;
```
sharanyavinod, line 88: Maybe reset busy in finally here so its not stuck in case the aem async throws?

nx2/blocks/canvas-actions/canvas-actions.js
```diff
@@ -1,19 +1,133 @@
...
+    this.requestUpdate();
```
sharanyavinod: Is this needed? Wouldnt change in busy already trigger an update?

nx2/blocks/canvas-actions/canvas-actions.js
```diff
@@ -1,19 +1,133 @@
...
+  static properties = {
+    _sendIcon: { state: true },
+    _busy: { state: true },
+    _error: { state: true },
+  };
+
+  constructor() {
+    super();
+    this._hash = new HashController(this);
+    this._busy = false;
```
sharanyavinod: Would be good to check if these are really needed since repo rules ask to avoid unnecessary constructors

nx2/blocks/canvas-actions/canvas-actions.js
```diff
@@ -1,19 +1,133 @@
...
+import { loadHrefSvg } from '../../utils/svg.js';
+import '../shared/popover/popover.js';
 
 const style = await loadStyle(import.meta.url);
 
+const SEND_ICON_HREF = new URL('../img/icons/S2_Icon_Send_20_N.svg', import.meta.url).href;
+
 class NXCanvasActions extends LitElement {
+  static properties = {
+    _sendIcon: { state: true },
```
sharanyavinod: icon never changes once loaded from what I see - could be just in a firstUpdated and cleaned up from properties

nx2/blocks/canvas/canvas.js
```diff
@@ -155,6 +163,11 @@ export default async function decorate(block) {
   const mountRoot = canvasEditorMountRoot(block);
   mountRoot.classList.add('nx-canvas-editor-mount');
 
+  mountRoot.addEventListener('nx-editor-undo-state', (e) => {
```
sharanyavinod, line 166: For future - maybe we should move all events out to constants and reuse so we have a clear view + avoid regressions

nx2/blocks/canvas-actions/canvas-actions.js
hannessolo: daFetch catches on the `fetch` call so this call shouldn't throw an exception.

nx2/blocks/canvas/nx-editor-doc/prose.js
hannessolo: Simplified this function.

nx2/blocks/canvas/canvas.js
hannessolo: Agree, this would be a good refactoring task.

sharanyavinod (approved):

PR #387, opened 2026-04-24
feat: setup tools panel

sharanyavinod:
- Refactors the panel system to remove the nx-panel-header intermediate layer and simplify panel open/close
- Introduces NxToolPanel (was a placeholder): managed shell with a consumer picker, lazy-loaded consumer content, first-party header-actions zone, and close button
- Updates chat to fully own the panel content including header with close
- Updates browse to have chat toggle similar to canvas

Preview: 
https://da.live/canvas?nx=ew-panels&nxver=2#/exp-workspace/frescopa/index
https://da.live/browse?nx=ew-panels&nxver=2#/aem-sandbox/block-collection/drafts

### DIFF
--- a/nx2/blocks/canvas/canvas.js
+++ b/nx2/blocks/canvas/canvas.js
@@ -1,5 +1,5 @@
 import { loadStyle, hashChange } from '../../utils/utils.js';
-import { hidePanel, unhidePanel, openPanelWithFragment } from '../../utils/panel.js';
+import { getPanelStore, openPanel } from '../../utils/panel.js';
 import './nx-canvas-header/nx-canvas-header.js';
 import './nx-editor-doc/nx-editor-doc.js';
 import './nx-editor-wysiwyg/nx-editor-wysiwyg.js';
@@ -12,11 +12,6 @@ function buildCanvasDocPath(state) {
   if (!org || !site || !path) return null;
   return `${org}/${site}/${path}`;
 }
-
-const FRAGMENTS = {
-  before: 'https://da.live/fragments/exp-workspace/chat',
-  after: 'https://da.live/fragments/exp-workspace/tool',
-};
 
 const CANVAS_EDITOR_VIEW_KEY = 'nx-canvas-editor-view';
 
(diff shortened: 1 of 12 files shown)


hannessolo:
Screenshot 2026-04-24 at 10 30 12

In the chat panel, the toggle button isn't at the edge of the panel anymore if it's wide enough

sharanyavinod:
> In the chat panel, the toggle button isn't at the edge of the panel anymore if it's wide enough

Oh nice catch! Will put in a fix

nx2/utils/panel.js
```diff
@@ -205,9 +208,22 @@ export async function restorePanels() {
...
+export async function openPanel({ position, width = '400px', getContent } = {}) {
+  const existing = document.querySelector(`aside.panel[data-position="${position}"]`);
+  if (existing && !existing.hidden) return existing;
+  if (existing?.hidden) {
+    showPanel(existing);
+    return existing;
+  }
+
+  const beforeMain = position === 'before';
+  const content = await getContent();
```
hannessolo: I think we could have a use case where we open a panel without `getContent` - not in our code but this panel is part of the general nx framework.

So would it make sense to have a check like `getContent?.()`?

nx2/blocks/canvas/canvas.js
```diff
@@ -99,41 +92,39 @@ function syncCanvasEditorsToHash({ mountRoot, header, state }) {
...
+    getContent: async () => {
+      const { loadFragment } = await import('../fragment/fragment.js');
+      await import('../tool-panel/tool-panel.js');
+      const toolPanel = document.createElement('nx-tool-panel');
+      toolPanel.consumers = [
+        {
+          id: 'tools',
+          label: 'Tools',
+          firstParty: false,
+          load: () => loadFragment('https://da.live/fragments/exp-workspace/tool'),
```
hannessolo: I'm not sure this load fragment makes sense anymore. If `canvas` owns the list of views shown in the tool panel, it can also just render them directly instead of having a fragment load them. So I think we can remove the whole fragment piece from canvas (keep it for tool-panel though, since other parts of the code could use it in the future).

Eg. we would just add to the `consumers` array the component for page outline that you're working on, and a generic `extensionLoader` component for all the configured extensions.

nx2/blocks/canvas/canvas.js
```diff
@@ -99,41 +92,39 @@ function syncCanvasEditorsToHash({ mountRoot, header, state }) {
...
+      return document.createElement('nx-chat');
+    },
+  },
+  after: {
+    width: '400px',
+    getContent: async () => {
+      const { loadFragment } = await import('../fragment/fragment.js');
+      await import('../tool-panel/tool-panel.js');
+      const toolPanel = document.createElement('nx-tool-panel');
+      toolPanel.consumers = [
```
hannessolo: I'm not sure `consumers` is a very clear name to me, maybe `views` or `children`?

nx2/blocks/canvas/canvas.js
sharanyavinod: I was thinking of doing that when introducing the content for right panel like outline, but you're right, might as well clean it up already

hannessolo (approved):

PR #390, opened 2026-04-24
fix: avoid using fragment panels

sharanyavinod:
Previous panel stored as fragment and current implementation are creating duplicates -> fix to avoid using previously stored fragments within canvas

Preview: https://da.live/canvas?nx=ew-panels&nxver=2#/exp-workspace/frescopa/index

### DIFF
--- a/nx2/blocks/canvas/canvas.js
+++ b/nx2/blocks/canvas/canvas.js
@@ -157,7 +157,7 @@ export default async function decorate(block) {
     syncCanvasEditorsToHash({ mountRoot, header, state });
   });
 
-  // Restore panels opened by canvas (no fragment URL). Fragment-based entries are legacy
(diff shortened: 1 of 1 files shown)


nx2/blocks/canvas/canvas.js
```diff
@@ -157,8 +157,9 @@ export default async function decorate(block) {
     syncCanvasEditorsToHash({ mountRoot, header, state });
   });
 
-  // Restore any panels that were open in the previous session.
+  // Restore panels opened by canvas (no fragment URL). Fragment-based entries are legacy
```
hannessolo: I'm not sure I'd call it legacy - other features that use nx might want to use panels with fragments, as that's a common edge delivery pattern. Here it makes more sense if canvas just controls them.

hannessolo (approved):

PR #391, opened 2026-04-24
[ew] implement breadcrumbs in header

hannessolo:
https://ew-breadcrumb--da-nx--adobe.aem.live

https://da.live/canvas?nx=ew-breadcrumb&nxver=2#/exp-workspace/frescopa/index
https://da.live/browse?nx=ew-breadcrumb&nxver=2#/exp-workspace/frescopa

### DIFF
--- a/nx2/blocks/shared/breadcrumb/breadcrumb.css
+++ b/nx2/blocks/shared/breadcrumb/breadcrumb.css
@@ -1,20 +1,9 @@
 :host {
   display: block;
   min-width: 0;
-  height: var(--s2-spacing-500);
-  box-sizing: border-box;
 }
 
-.breadcrumb {
-  display: flex;
-  align-items: center;
-  min-width: 0;
-  height: 100%;
-  overflow: hidden;
-  box-sizing: border-box;
-}
-
-.breadcrumb ol {
+.nx-breadcrumb ol {
   display: flex;
   align-items: center;
   flex-wrap: nowrap;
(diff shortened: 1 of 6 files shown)


kozmaadrian (approved):

nx2/blocks/shared/breadcrumb/breadcrumb.css
```diff
@@ -1,34 +1,30 @@
...
   font-weight: var(--s2-component-m-regular-font-weight);
   color: var(--s2-gray-700);
 }
 
+:host([variant="large"]) .nx-breadcrumb ol {
+  font-size: var(--s2-component-l-regular-font-size);
+  line-height: var(--s2-component-l-regular-line-height);
+  font-weight: var(--s2-component-l-regular-font-weight);
+}
+
```
sharanyavinod: The styles reliant on :host([variant="large"]) -> could we have them nested within :host? More a nit, but would be nice if related styles are together.

sharanyavinod (approved):
Font size as per mocks is 14px, whereas currently its 16px. Intentional? Looks a bit inconsistent with the rest of the UI atm.
Other than that, looks good.

hannessolo:
> Font size as per mocks is 14px, whereas currently its 16px. Intentional?

Sort of intentional? I somehow though we wanted the one in the shell to be bigger than the one in the browse view. But you're right, it's 14px in both in the figma.

Let me remove the large styling variant then. Anyway, I'm still of the opinion that everything should be bigger...

sharanyavinod:
@hannessolo Indeed it should be everything or nothing. Right now, its just a bit off since the others are small and breadcrumb grabs more attention.

PR #393, opened 2026-04-27
feat: Implements the action bar with core actions.

kozmaadrian:
Implements the action bar with core actions: preview, publish, delete, and rename.

Also adds shared components dialog, toast, and progress circle, required for the browse actions.

Test URLs:
- https://da.live/browse?nx=ew-browseactions&nxver=2#/exp-workspace/frescopa


hannessolo:
Screenshot 2026-04-28 at 10 38 57

The action buttons in the modal have the wrong font family (maybe the variable isn't set there for some reason?)

hannessolo:
I think we should separate preview and publish as actions. As an edge delivery user, that's what I'm used to. Only having a publish button makes it not clear to me whether it will do preview, publish or preview -> publish both.

hannessolo:
Happy to debate this one, but: Do we really need a confirmation dialog for single-file preview/publish? For multi files it makes sense, but for single file, we don't have a confirmation in sidekick or da so I don't think we need it here.

sharanyavinod:
Screenshot 2026-04-28 at 10 54 21
Currently if the url has a hash or a query param, that is also part of the browse title and breadcrumb. Would be good to sanitize so these stay meaningful.

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,173 @@
...
+    body: { type: Object, attribute: false },
+    primaryActionLabel: { type: String, attribute: false },
+    primaryActionId: { type: String, attribute: false },
+    primaryActionDisabled: { type: Boolean, attribute: false },
+    cancelLabel: { type: String, attribute: false },
+    /** When **`VARIANT_DESTRUCTIVE`**, primary uses danger styling. */
+    variant: { type: String, attribute: false },
+    autofocusId: { type: String, attribute: false },
+    dismissable: { type: Boolean, attribute: false },
+    onPrimaryAction: { type: Function, attribute: false },
```
sharanyavinod: Why do we go for this approach vs firing events as in other blocks? This creates tight coupling between consumer and dialog, when ideally it would be nice that dialog simply fires an event and consumer listens and responds - consistent with how things are working across the repo atm. Dialog doesnt and shouldnt know about what happens after click - it just handles the UI, fires event and forgets, and everything else is the responsibility of the consumer.
As a bonus, this would also help redice the properties ie both the onAction and the actionId

sharanyavinod:
@kozmaadrian All new blocks introduced have the callback pattern as in React vs event driven pattern across the repo. Why this deviation from pattern?

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,173 @@
...
+        id=${primaryActionId || nothing}
+        ?disabled=${Boolean(primaryActionDisabled)}
+        @click=${onPrimaryAction}
+      >${this.primaryActionLabel}</button>
+    `;
+  }
+
+  _renderShell({ title, body, actions }) {
+    return html`
+      <div
```
sharanyavinod: Worth exploring if we can use the native <dialog> api here instead of custom code completely - it would give ootb support for many of the expected capabilities of a dialog.

hannessolo: +1, this should use `dialog` with `showModal`

nx2/blocks/browse/action-bar/action-bar.js
```diff
@@ -0,0 +1,117 @@
...
+const styles = await loadStyle(import.meta.url);
+
+export class NxBrowseActionBar extends LitElement {
+  static properties = {
+    count: { type: Number },
+    showDelete: { type: Boolean },
+    showRename: { type: Boolean },
+    showDeploy: { type: Boolean },
+    disabled: { type: Boolean, attribute: false },
+    onDismiss: { type: Function, attribute: false },
```
anfibiacreativa: onDismiss/onAction are callback props. since we use CustomEvent, why don't we fire events instead? same for onComplete in the action dialogs

nx2/blocks/browse/actions/deploy/deploy.js
```diff
@@ -0,0 +1,154 @@
...
+        },
+      };
+    }
+    if (phase === 'preview') {
+      const url = r.json?.preview?.url;
+      if (url) openAemUrlWithNoCache(url);
+    } else {
+      const url = r.json?.live?.url;
+      if (url) openAemUrlWithNoCache(url);
+    }
```
anfibiacreativa, line 40: openAemUrlWithNoCache opens tabs inside the deploy fn. why don't we return the URLs and let the consumer decide? (in the function in line 8)

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,173 @@
...
+const styles = await loadStyle(import.meta.url);
+
+/** Primary button uses danger styling (browse delete). */
+export const VARIANT_DESTRUCTIVE = 'destructive';
+
+class NxDialog extends LitElement {
+  static properties = {
+    title: { type: String, attribute: false },
+    body: { type: Object, attribute: false },
+    primaryActionLabel: { type: String, attribute: false },
```
anfibiacreativa: I see the same issue here with function props instead of events 
https://lit.dev/docs/components/events/

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,173 @@
...
+  _renderShell({ title, body, actions }) {
+    return html`
+      <div
+        class="scrim"
+        @click=${this._onBackdropClick}
+      >
+        <div
+          class="panel"
+          role="dialog"
+          aria-modal="true"
```
anfibiacreativa: we have role="dialog" aria-modal="true" but no focus trap. tab will escape to background.

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,173 @@
...
+      el.focus();
+      if (el instanceof HTMLInputElement) {
+        el.select();
+      }
+    });
+  }
+
+  _onLayerClose = () => {
+    this.dispatchEvent(
+      new CustomEvent('nx-dialog-close', { bubbles: true, composed: true }),
```
anfibiacreativa, line 68: when the dialog closes, focus needs to return to whatever triggered it.

nx2/blocks/browse/action-bar/action-bar.js
```diff
@@ -0,0 +1,117 @@
...
+    const svg = this._icons?.[iconKey];
+    return svg ? svg.cloneNode(true) : nothing;
+  }
+
+  render() {
+    const n = this.count ?? 0;
+    const label = n === 1 ? '1 item selected' : `${n} items selected`;
+
+    return html`
+      <div class="bar" role="toolbar" aria-label="Folder selection">
```
anfibiacreativa: aria-label="Folder selection" but this toolbar handles files too, right?

nx2/blocks/browse/browse.js
```diff
@@ -20,10 +25,23 @@ const documentLayoutStyles = await loadStyle(
 );
 document.adoptedStyleSheets = [...document.adoptedStyleSheets, documentLayoutStyles];
 
+/** @param {{ title?: string, body?: string }} m */
+function browseActionErrorToastText(m) {
```
anfibiacreativa: this is a pure string transform, why don't we move it to nx2/utils/, as we have moved other utilities

nx2/blocks/browse/browse-api.js
```diff
@@ -1,18 +1,7 @@
...
- */
 export async function listFolder(fullpath) {
-  let response;
-  try {
-    response = await daFetch(`${DA_ORIGIN}/list${fullpath}`);
-  } catch (err) {
-    const message = err instanceof Error ? err.message : 'List request failed';
-    return { error: message, status: 0 };
-  }
+  const response = await daFetch(`${DA_ORIGIN}/list${fullpath}`);
```
anfibiacreativa: the try/catch around daFetch in listFolder was removed. if daFetch throws (network error), this is now unhandled. is that intentional?

nx2/blocks/browse/utils.js
```diff
@@ -50,20 +37,25 @@ export const ICONS = {
...
+  close: `${ICONS_BASE}S2_Icon_Close_20_N.svg`,
+  delete: `${ICONS_BASE}S2_Icon_Delete_20_N.svg`,
+  edit: `${ICONS_BASE}S2_Icon_Edit_20_N.svg`,
+  preview: `${ICONS_BASE}S2_Icon_Preview_20_N.svg`,
+  publish: `${ICONS_BASE}S2_Icon_Publish_20_N.svg`,
+  rename: `${ICONS_BASE}S2_Icon_Rename_20_N.svg`,
 };
 
 export function entryTypeFromExtension(ext) {
   if (ext == null || ext === '') {
```
anfibiacreativa: it seems it no longer calls .toLowerCase() or strips leading dots. if an extension comes in as .HTML, this silently breaks. why remove it?

nx2/blocks/shared/toast/toast.css
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    & .close-icon {
+      display: block;
+      flex-shrink: 0;
+      width: 20px;
+      height: 20px;
+      background: currentcolor;
+      mask-image: url("/nx2/img/icons/S2_Icon_Close_20_N.svg");
+      mask-repeat: no-repeat;
```
anfibiacreativa, line 84: I think we use loadHrefSvg or import.meta.url-relative. to avoid absolute paths that break if assets move.

nx2/blocks/browse/action-bar/action-bar.css
```diff
@@ -0,0 +1,134 @@
...
+  line-height: var(--s2-component-m-regular-line-height);
+  font-weight: var(--s2-component-m-medium-font-weight);
+  color: var(--s2-gray-900);
+  background-color: var(--s2-gray-25);
+}
+
+.bar-lead {
+  display: flex;
+  align-items: center;
+  gap: 6px;
```
anfibiacreativa: magic number. please let's move to var

nx2/blocks/browse/action-bar/action-bar.css
```diff
@@ -0,0 +1,134 @@
...
+  box-sizing: border-box;
+
+  /* Fixed width (viewport-based only) so actions do not shift when the count string changes. */
+  flex-grow: 0;
+  flex-shrink: 0;
+
+  --bar-lead-w: 141px;
+
+  width: var(--bar-lead-w);
+  min-width: var(--bar-lead-w);
```
anfibiacreativa: width, min-width, max-width all the same var. flex: 0 0 var(--bar-lead-w) does the job.

nx2/blocks/browse/action-bar/action-bar.css
```diff
@@ -0,0 +1,134 @@
...
+
+.bar {
+  display: flex;
+  align-items: center;
+  gap: var(--s2-spacing-200);
+  box-sizing: border-box;
+  width: 100%;
+  min-height: var(--s2-spacing-500);
+  height: var(--s2-spacing-500);
+  padding-left: 10px;
```
anfibiacreativa: magic number

nx2/blocks/browse/actions/delete/delete.css
```diff
@@ -0,0 +1,30 @@
+:host {
+  display: contents;
+}
+
+.browse-action-root {
+  display: contents;
+}
+
+.browse-action-busy {
+  position: fixed;
```
anfibiacreativa: isn't this identical in delete.css, deploy.css, rename.css. why don't we extract it?

nx2/blocks/shared/dialog/dialog.css
```diff
@@ -0,0 +1,242 @@
...
+    width: 100%;
+    height: 0;
+    border: 0;
+    border-top: 1px solid var(--s2-gray-200);
+  }
+
+  & .title {
+    margin: 0;
+    padding: 0;
+    font-size: 18px;
```
anfibiacreativa: magic number

anfibiacreativa (changes requested):

anfibiacreativa:
I would really love for us to establish the no single letter name var and the boolean is/has prefix, but I won't block a PR for this. however there are many instances of those potential changes, and one of `handleEventName` instead of `onEventName` that I really stylistically prefer

nx2/blocks/browse/actions/deploy/deploy.js
kozmaadrian: Done as suggested.

nx2/blocks/browse/action-bar/action-bar.js
kozmaadrian: I can changed that, however at one point, the agreement was to consistently use events in shared components, while minimizing their use in more tightly coupled scenarios.

nx2/blocks/shared/toast/toast.css
kozmaadrian: This pattern is used across different components, and I think it makes a lot more sense to follow the same approach here..

nx2/blocks/browse/action-bar/action-bar.css
kozmaadrian: The magic number seemed more readable since there’s no --s2-* variable defined for 6px.

nx2/blocks/browse/action-bar/action-bar.css
kozmaadrian: Same here.

nx2/blocks/browse/actions/delete/delete.css
kozmaadrian: Refactored.

nx2/blocks/browse/browse-api.js
kozmaadrian: daFetch already catches all the errors, so it was redundant here.

nx2/blocks/browse/actions/deploy/deploy.js
```diff
@@ -0,0 +1,107 @@
...
+      const url = saveResult.json?.live?.url;
+      if (url) openedUrls.push(url);
+    }
+  }
+  return { ok: true, openedUrls };
+}
+
+const styles = await loadStyle(import.meta.url);
+
+class NxBrowseDeployRunner extends LitElement {
```
sharanyavinod, line 47: I'm not sure I understand why we need this component? Any reason why this logic is separate and not part of browse? For me this is overkill and feels antipattern.

nx2/blocks/shared/overlay/overlay.js
```diff
@@ -0,0 +1,29 @@
+import { LitElement, html } from 'da-lit';
+import { loadStyle } from '../../../utils/utils.js';
+
+const styles = await loadStyle(import.meta.url);
+
+class NxOverlay extends LitElement {
```
sharanyavinod: Since this is only used in browse I would move this out of shared for now. Without finalized mocks and given how minimal it is, I think its premature to keep it in shared - ideally it neednt be a component at all and could just be done inline or we move it into browse.

nx2/blocks/browse/actions/deploy/deploy.js
kozmaadrian: I split it out to match the other actions and keep browse simpler. It also leaves room if we add multi-item preview/publish later.

nx2/blocks/shared/overlay/overlay.js
kozmaadrian: Fair! I’ll move the overlay into browse so it’s not in shared anymore. We can inline it later if we want it even simpler.

nx2/blocks/shared/dialog/dialog.js
kozmaadrian: The dialog was refactored, no callbacks are used.

nx2/blocks/browse/action-bar/action-bar.js
kozmaadrian: Update: AFAIU, reducing the number of events can be addressed in a future refactoring if needed. Therefore, I updated the code to rely exclusively on events.

nx2/blocks/shared/dialog/dialog.js
kozmaadrian: The dialog has been refactored.

nx2/blocks/shared/dialog/dialog.js
kozmaadrian: Switched to <dialog> element.

nx2/blocks/browse/actions/deploy/deploy.js
sharanyavinod: I get the need to keep things aligned, but consistency for the sake of consistency isnt helpful either. Delete and rename are valid as components because they have actual interactive state and user interaction - this has neither. In such a case, I feel a Lit component isnt justified. Completely fair to revisit and adapt when we have a more concrete use case, but not atm.

PR #394, opened 2026-04-27
media-library(refactor) : decoupling indexing and display  

kmurugulla:
Decouples media library indexing from display using event-based architecture and web workers, enabling
  plugin mode (indexing in background) and app mode (UI-driven indexing).

  Key Changes:
  - Move indexing to web workers (full & incremental builds run off main thread)
  - Event-based communication between indexing and display layers
  - New dual polling strategy: app mode polls index files, plugin mode polls status API
  - Consolidate UI layer into ui/ directory, indexing logic into indexing/worker/

Test URLs:
- After: https://da.live/apps/media-library?nx=mldecouple&debug=perf#/aemsites/vitamix


amol-anand (commented):
Overall great idea to de couple the indexing bit from the viewing bit so we can keep those two functionalities separated out but still do everything in the browser for now. Found some stuff with Claude's help that might need some refactoring. But outside of that it looks cleanly separated out.

nx/blocks/media-library/core/constants.js
```diff
@@ -109,14 +109,8 @@ export const ExternalMedia = Object.freeze({
...
-  if (href.includes('localhost')) return 'http://localhost:8787';
-  return 'https://da-etc.adobeaem.workers.dev';
-}
-
-export const DA_ETC_ORIGIN = getDaEtcOrigin();
+// Worker-safe constants (duplicated here to avoid importing from public/utils/constants.js
+// which has window.location access at module level)
+export const AEM_ORIGIN = 'https://admin.hlx.page';
+export const DA_ORIGIN = 'https://admin.da.live';
+export const DA_ETC_ORIGIN = 'https://da-etc.adobeaem.workers.dev';
```
amol-anand: 🔴 **Blocker — env regression.** These origins are now static prod URLs. On `main`, `DA_ORIGIN` resolved dynamically via `nx/public/utils/constants.js::getDaEnv` (`da-admin` URL param / localStorage → `local | stage | prod`), and `DA_ETC_ORIGIN` had its own `da-etc` switch. Every consumer in this PR — `indexing/admin-api.js`, `indexing/locks.js`, `indexing/build.js` (which then ships `daOrigin`/`daEtcOrigin` into the worker `context`), `ui/data.js`, `core/paths.js`, `core/urls.js::etcFetch` — now imports from here, so opening the media library with `?da-admin=stage` or `?da-admin=local` still talks to prod. Stage/local development is effectively offline.

The "worker-safe" justification is fine *for the worker boundary*. Suggested fix:

1. Have `core/constants.js` re-export `DA_ORIGIN` / `AEM_ORIGIN` from `public/utils/constants.js` for main-thread code (matches the previous import path).
2. Restore the `getDaEtcOrigin()` lookup (you removed it in this hunk) for `DA_ETC_ORIGIN`.
3. In `indexing/build.js::runWorkerBuild`, resolve those values once on the main thread and continue passing them in `postMessage`. The worker already takes `daOrigin`/`daEtcOrigin` from `context`, so no worker change is needed.

nx/blocks/media-library/media-library.js
```diff
@@ -620,8 +621,28 @@ class NxMediaLibrary extends LitElement {
...
+        // App policy: Auto-trigger builds when index is missing (app mode only)
+        if (event.type === 'index-missing' && isApp) {
+          const [siteOrg, siteRepo] = this.sitePath.split('/').slice(1, 3);
+          triggerBuild(this.sitePath, siteOrg, siteRepo);
+        }
+      };
+
+      const hasMediaData = (getAppState().mediaData?.length || 0) > 0;
+
+      initService(this.sitePath, { onEvent, mode, hasMediaData });
```
amol-anand, line 646: 🔴 **Blocker — dead destructure / broken UI states.** A few lines below (head lines 688–713, unchanged in this PR), `loadMediaData` still does:

```js
const { data, indexMissing, indexing, lockFresh } = await loadMediaSheet(...);
if (indexing) { /* lock-held branch */ return; }
...
isBackgroundRefreshInProgress: !!(data?.length > 0) && !!lockFresh,
```

But `ui/data.js::loadMediaSheet` only returns `{ data }` or `{ data, indexMissing }` now — `indexing` and `lockFresh` are `undefined`. So the lock-held branch is **dead** and `isBackgroundRefreshInProgress` is **always `false`** on initial load.

The new `LOCK_DETECTED` event in `coordinator.initService` (called here on this line) does eventually fix the state, but only after its async lock check resolves — and `loadMediaData` has already awaited and possibly cleared the UI by then, producing a flash of empty state in the lock-held case.

Suggested fix: parallelise `coordinator.initService`'s lock check with `loadMediaData` (or run it first) so `LOCK_DETECTED` lands before/with `setMediaData`. Then **delete the dead `indexing` / `lockFresh` destructure and the `if (indexing)` block** (head L688–713) — they're misleading.

Also note: `loadMediaSheet` is now called twice on first paint — once here in `loadMediaData`, then again at `coordinator.js:415` when `!hasMediaData && !freshLock`. That redundant `index.json` fetch should be folded into the existing one.

nx/blocks/media-library/indexing/worker/linked-content.js
```diff
@@ -0,0 +1,183 @@
...
+      const { pages: linkedPages, firstDiscoveredTimestamp } = data;
+      linkedPages.forEach((doc) => {
+        const entry = toExternalMediaEntry(url, doc, firstDiscoveredTimestamp, org, repo);
+        if (entry) {
+          updatedIndex.push(entry);
+          added += 1;
+        }
+      });
+    });
+  }
```
amol-anand: 🟠 **Correctness — external-media entries duplicate forever in incremental builds.** This block unconditionally `push`es a `(url, doc)` entry per linked page. `processLinkedContent` is also called from `worker/incremental.js:437` (with `prebuiltUsageMap = null`) where `updatedIndex` starts as the **previous** index. So each incremental build re-adds every external-URL/page pair → unbounded growth, and stale references never get removed when a page stops linking the URL.

Compare with the canonical (non-worker) path in `indexing/linked-content.js:350–388`, which:
1. `purgeInvalidExternalMediaEntries(updatedIndex)` first.
2. Computes `obsolete = updatedIndex.filter(isExtlinksEntry && (e.doc === '' || !linkedPages.includes(e.doc)))` and splices them.
3. For each `(url, doc)`, finds `existingIdx` and **updates** rather than pushes; only pushes when `existingIdx === -1`.
4. Runs `validateExternalMediaEntries` HEAD-checks afterward.

The worker version is missing all of that. Please port the dedupe + obsolete-purge logic (or share a single implementation between the two paths) and add a unit test that runs `processLinkedContent` twice over the same `updatedIndex` and asserts no growth.

nx/blocks/media-library/indexing/build.js
```diff
@@ -1,981 +1,267 @@
...
-    emitEarlyLinked();
+    const result = await resultPromise;
+    return result;
   } finally {
-    await medialogPromise?.catch(() => {});
-    await auditlogPromise?.catch(() => {});
+    // Clean up
+    worker.terminate();
+    URL.revokeObjectURL(workerBlobUrl);
   }
```
amol-anand, line 201: 🟡 **No watchdog / hang protection.** If the worker deadlocks or its `onmessage` throws inside an `await` without posting a `type: 'error'` message, `resultPromise` never settles → the `try`/`finally` here never runs → `worker.terminate()` is skipped → `triggerBuild`'s outer `finally` never resumes polling. The whole indexing pipeline is wedged until the page reloads.

Suggested: add a watchdog timer (e.g. `IndexConfig.STATUS_POLL_MAX_DURATION_MS` or a new `BUILD_MAX_DURATION_MS`) that rejects the promise and force-terminates the worker. Heartbeat-style progress events from the worker could also be used to extend the deadline.

nx/blocks/media-library/indexing/build.js
```diff
@@ -1,981 +1,267 @@
...
+          worker.postMessage({ type: 'token-refresh-response', requestId, token: freshToken });
+        } catch (err) {
+          worker.postMessage({ type: 'token-refresh-response', requestId, token: null, error: err.message });
+        }
+      } else if (type === 'success') {
+        resolve(data);
+      } else if (type === 'error') {
+        reject(new Error(error.message || 'Worker error'));
       }
+    };
```
amol-anand: 🟡 **`onmessage` is `async` — message ordering can interleave.** A long `getAemSiteToken()` call inside the `'token-refresh'` branch yields the message-loop, so a subsequent `'progressive'` / `'success'` / `'error'` message can be processed before the refresh resolves. Today they don't share state but it's a foot-gun.

Also, in `worker/fetch.js::requestTokenRefresh`, every refresh request adds its own `message` listener and removes only its own. Two concurrent refreshes stack listeners. Suggested: a single shared listener + `Map<requestId, resolver>` (or replace with a `MessageChannel` per request).

nx/blocks/media-library/indexing/coordinator.js
```diff
@@ -1,168 +1,248 @@
...
+  createLockDetectedEvent,
+  createIndexMissingEvent,
+  createIndexLoadedEvent,
+} from './events.js';
+
+const CONFIG = {
+  INWARD_POLLING_INTERVAL: 60000,
+  OUTWARD_POLLING_INTERVAL: 120000,
+  LOCK_CHECK_INTERVAL: 5000,
+};
```
amol-anand: 🟡 **Move polling intervals into `IndexConfig`.** These constants belong next to `STATUS_POLL_INTERVAL_MS`, `LOCK_HEARTBEAT_INTERVAL_MS`, `LOCK_STALE_THRESHOLD_MS` in `core/constants.js::IndexConfig` — keeps timing knobs in one place and lets tests override them.

amol-anand (commented):
Thanks for the substantial refactor — the worker-based indexing + neutral event bus + UI/indexing folder split is a clear improvement, and the new `events.js` / `core/indexing-adapter.js` boundary is clean. `npm run lint:js` introduces zero new errors in media-library, and the existing 162 tests still pass.

Before merge, two regressions need to be addressed (🔴 inline below), plus one correctness bug in worker incremental builds. A handful of smaller items are flagged as suggestions.

### 🔴 Blockers
1. **Hardcoded `DA_ORIGIN` / `DA_ETC_ORIGIN` / `AEM_ORIGIN`** in `core/constants.js` break stage & local environments for every media-library API call (admin, locks, paths, ui/data, build's worker context). Restore dynamic resolution on the main thread; only freeze the value when handing it to the worker. (See inline on `core/constants.js`.)
2. **Dead `indexing` / `lockFresh` destructure** in `media-library.js::loadMediaData` — `ui/data.js::loadMediaSheet` no longer returns those keys, so `if (indexing) { … }` is unreachable and `isBackgroundRefreshInProgress: !!(data?.length > 0) && !!lockFresh` is always `false`. The new `LOCK_DETECTED` event eventually fixes the state, but only after `coordinator.initService`'s async lock check completes. (See inline on `media-library.js`.)

### 🟠 Correctness
3. **External-media duplicates forever in incremental builds** — the worker's `linked-content.js` unconditionally `push`es entries without dedupe or obsolete-purge. (See inline.)

### 🟡 Suggestions (also inline)
- Worker has no watchdog timeout; a hung worker hangs the indexing pipeline.
- `worker.onmessage` is `async` and shares a single token-refresh listener — concurrent refreshes can stack listeners.
- `coordinator.js`: `pollingStarted` module-global is sticky across mode switches; `eventEmitter` is overwritten without unsubscribing previous handlers.
- `coordinator.js`: control-flow via `error.message?.includes('Index build already in progress')` — better as a typed error code.
- `coordinator.js`: hardcoded polling intervals belong in `IndexConfig` next to their cousins.
- `core/indexing-adapter.js`: redundant duplicate `updateAppState({ persistentError })` in `BUILD_ERROR` handler.

### Other small things
- `coordinator.js::triggerBuild` builds the `LOCK_DETECTED` event by hand instead of via `createLockDetectedEvent(...)`.
- `core/urls.js::etcFetch` (top-level, main-thread) and `worker/fetch.js::etcFetch` (per-call origin) have diverged; consolidate after fix #1.
- No worker tests — please add at least: external-media dedupe (regression for #3), `indexing-adapter` event→state, and a `coordinator` lock-detected→poll happy path.

Happy to re-review after #1–#3 are addressed.

nx/blocks/media-library/indexing/coordinator.js
```diff
@@ -338,100 +343,94 @@ export async function triggerBuild(sitePath, org, repo, ref = 'main') {
...
   const [org, repo] = sitePath.split('/').slice(1, 3);
 
+  // Plugin mode: Load data once, no polling
+  if (mode === 'plugin') {
+    if (isPerfEnabled()) {
+      // eslint-disable-next-line no-console
+      console.log(`[perf] Initializing indexing service in PLUGIN mode (no polling) for ${sitePath}`);
+    }
+    return;
+  }
```
amol-anand, line 409: 🟡 **Module-level state is sticky across mode switches.** `pollingStarted`, `eventEmitter`, `inwardPollingInterval`, etc. are module globals. If the page first mounts in `plugin` mode and later in `app` mode (or the component remounts on hash navigation), `pollingStarted` short-circuits the second `initService` before reaching the polling block — but `eventEmitter` is overwritten silently, so a stale component instance can keep receiving events.

Suggested: tie state to a per-service object (return a `{ dispose }` handle from `initService`), and key any "already started" check on `(sitePath, mode)` rather than a boolean. At minimum, in `plugin` mode (L382-388), don't return without setting some "plugin initialised" flag, otherwise a later `app`-mode init will silently no-op against the original `eventEmitter`.

nx/blocks/media-library/indexing/coordinator.js
```diff
@@ -265,60 +305,25 @@ export async function triggerBuild(sitePath, org, repo, ref = 'main') {
...
-        progressiveCountCapped: false,
-        persistentError: null,
+    emit(createBuildCompleteEvent(
+      result.mediaData || [],
+      duration,
+      result.hasChanges,
+      result.lockRemoveFailed,
+    ));
+  } catch (error) {
+    if (error.message?.includes('Index build already in progress')) {
```
amol-anand: 🟡 **String-sniff control flow.** `error.message?.includes('Index build already in progress')` is brittle — any future tweak to the message in `build.js:195-197` silently breaks the `LOCK_DETECTED` branch. Better to throw a typed `MediaLibraryError(ErrorCodes.LOCK_HELD_BY_OTHER, ...)` from `build.js` and check `error.code` here.

Small extra: a few lines below (L319–325), the `LOCK_DETECTED` event is constructed by hand instead of via `createLockDetectedEvent(...)` — please use the factory for consistency with the rest of the file.

nx/blocks/media-library/core/indexing-adapter.js
```diff
@@ -0,0 +1,264 @@
...
+        const userMessage = getErrorMessage(event.code, event.message);
+        showNotification(t('NOTIFY_ERROR'), userMessage, 'danger');
+      } else if (event.code === IndexingErrorCode.AUTH_REQUIRED) {
+        showNotification(t('NOTIFY_ERROR'), t('NOTIFY_SIGN_IN'), 'danger');
+      }
+
+      // Special handling for persistent errors
+      if (event.isPersistent) {
+        updateAppState({ persistentError: { message: event.message } });
+      }
```
amol-anand: 🟡 **Dead duplicate write.** A few lines above (L207-212) `updateAppState(updates)` already wrote `persistentError: { message: event.message }` when `event.isPersistent` was true. This block re-writes the same value. Safe to delete.

amol-anand (commented):
Thanks for the iteration. Reviewing the 6 commits since `b61c9747` (`20f91ca6`, `f1c69346`, `a65cbb36`, `32c999af`, `72b7f496`, `fc10fbec`).

### ✅ Resolved

| # | Item | Status |
|---|---|---|
| 1 | Hardcoded origins → env regression | ✅ `resolveDaOrigin/resolveDaEtcOrigin/resolveAemOrigin` in `core/constants.js`; `locationData` passed into worker; worker re-resolves via `worker.js`. |
| 2 | Dead `indexing`/`lockFresh` destructure | ✅ Removed; `initService` now runs in parallel with `loadMediaData` (clean comment explains why). |
| 3 | External-media duplicates in incremental | ✅ `purgeInvalidExternalMediaEntries` + obsolete-purge + `existingIdx` lookup; matches the canonical non-worker path. |
| 4 | Worker watchdog | ✅ `BUILD_MAX_DURATION_MS` (30 min) timer; cleared on `success`/`error`/`onerror`. |
| 5 | Async `onmessage` listener stacking | ✅ `worker/fetch.js` now uses a shared listener + `pendingTokenRefreshes` Map. |
| 9 | Dead duplicate `updateAppState({ persistentError })` | ✅ Removed from `core/indexing-adapter.js`. |
| – | Typed `LOCK_HELD_BY_OTHER` | ✅ `build.js` now throws `MediaLibraryError(ErrorCodes.LOCK_HELD_BY_OTHER, ...)` — but coordinator.js still detects via string match (see inline). |

Media-library tests pass (162/162); lint clean for this PR (the 23 pre-existing `getSvg` errors are unrelated).

### ❌ Claimed in commit message but not actually fixed

The `20f91ca6` commit body lists several fixes that didn't land — flagging inline:

1. **Polling intervals → IndexConfig** — the new `INDEX_POLLING_INTERVAL_MS` / `LOGS_POLLING_INTERVAL_MS` / `LOCK_CHECK_INTERVAL_MS` constants were added but `coordinator.js` still uses its local `CONFIG` block. The new IndexConfig values are dead.
2. **Replaced string error detection with typed `MediaLibraryError`** — `build.js` was updated to throw the typed error, but the consumer in `coordinator.js::triggerBuild` still does `error.message?.includes('Index build already in progress')`.
3. **`createLockDetectedEvent` factory** — still hand-built next to the string-sniff.
4. **Sticky module state (`pollingStarted`, `eventEmitter`)** — unchanged. Still a single module global; `eventEmitter` is silently overwritten on each `initService` call.

### ⚠️ New regressions / concerns introduced

- **`?da-admin=reset` no longer resets** and **`?da-admin=stage` no longer persists** in localStorage. The new `resolveDaOrigin` reads but never writes/clears localStorage, while the legacy `getDaEnv` in `nx/public/utils/constants.js` does both. (See inline.)
- **Main-thread vs worker resolution can disagree.** Worker has no `localStorage`, so a user with a *legacy* localStorage `da-admin=stage` (set by other parts of nexter) and *no* URL param will hit stage on the main thread but prod inside the worker. Cleanest fix: resolve once on the main thread, pass the resolved origin string to the worker (the `locationData` round-trip is what re-introduced the divergence).
- **`isMediaLibraryPluginMode()` called multiple times per render** in `mediainfo.js:788–797`. Minor, but each call re-reads `window.location.pathname`. `grid.js` already memoises into a `pluginMode` const at the top of the render — please mirror that pattern in `mediainfo.js`.
- **Hardcoded `'Copy'` string** in `mediainfo.js:797` (`${pluginMode ? t('UI_INSERT_BUTTON') : 'Copy'}`) bypasses i18n. Add a `UI_COPY_BUTTON` message and use it.
- **`getDaSdkActions` requires `actions.sendText`** (`export.js:67`) but `sendText` is never invoked since `a65cbb36` (link inserts switched to `sendHTML`). The check will reject on SDKs that only expose `sendHTML`. Drop `sendText` from the precondition.
- **`media-library.js:700` unconditionally clears `indexLockedByOther: false`** after `loadMediaSheet` resolves. If `LOCK_DETECTED` arrived first (lock check is now parallel — by design), this clobbers it back to `false`. Mostly cosmetic since the next poll re-detects, but the whole reason `initService` was moved earlier was to surface lock state before data arrives. (See inline.)
- **Watchdog never resets on activity.** A long-but-actively-progressing build is killed at 30 min. Resetting on `'progress'` / `'progressive'` messages would make this much safer for huge sites without raising the cap. Optional.
- **`getMediaLibraryAppHref` hardcodes `https://da.live/apps/media-library`** — `?nx=local`, `?da-admin=stage`, etc. are dropped from the deep link. Probably acceptable for production users but breaks dev/stage flows.
- **No tests added** for any of the new code paths (env resolver, watchdog, refresh-listener Map, external-media dedupe, plugin SDK insert, `result.silent` branch, plugin-mode UI labels). Given the scope of the new attack surface, please add at least a few — happy to enumerate priorities if helpful.

### Plugin-mode (new feature) commits — high level

- The `insertMediaViaPluginSdk` path with `actions.sendHTML` + `actions.closeLibrary?.()` looks reasonable. Returning `{ silent: true }` and gating the notification at the call sites is clean.
- The fallback chain (plugin SDK → rich clipboard → plain `writeText`) with distinct messages (`NOTIFY_PLUGIN_FALLBACK_*`) is a nice touch.
- The `INDEX_MISSING_PLUGIN` empty state with a deep link to the standalone app is good UX, modulo the env-loss above.
- The `tryClosePluginPanel` round-trip in `72b7f496` → reverted in `fc10fbec` reads cleanly — the link click now just opens the new tab without trying to also close the iframe.

Overall the architecture-level fixes (origin resolution, dedupe, watchdog, listener) are solid. The remaining items are mostly cleanup of half-applied changes in `coordinator.js` and a couple of small env-resolution edge cases.

nx/blocks/media-library/indexing/coordinator.js
```diff
@@ -1,168 +1,259 @@
...
+  createLockDetectedEvent,
+  createIndexMissingEvent,
+  createIndexLoadedEvent,
+} from './events.js';
+
+const CONFIG = {
+  INWARD_POLLING_INTERVAL: 60000,
+  OUTWARD_POLLING_INTERVAL: 120000,
+  LOCK_CHECK_INTERVAL: 5000,
+};
```
amol-anand: 🔁 **Half-applied.** `20f91ca6` added `INDEX_POLLING_INTERVAL_MS` / `LOGS_POLLING_INTERVAL_MS` / `LOCK_CHECK_INTERVAL_MS` to `IndexConfig` (constants.js L22–24), but this local `CONFIG` block is still here and is what `setInterval` actually consumes (L113, L171, L251). The new IndexConfig values are dead code today.

Quick fix: delete this block and replace `CONFIG.INWARD_POLLING_INTERVAL` → `IndexConfig.INDEX_POLLING_INTERVAL_MS`, `CONFIG.OUTWARD_POLLING_INTERVAL` → `IndexConfig.LOGS_POLLING_INTERVAL_MS`, `CONFIG.LOCK_CHECK_INTERVAL` → `IndexConfig.LOCK_CHECK_INTERVAL_MS`.

nx/blocks/media-library/indexing/coordinator.js
```diff
@@ -265,60 +316,25 @@ export async function triggerBuild(sitePath, org, repo, ref = 'main') {
...
+      emit({
+        type: IndexingEventType.LOCK_DETECTED,
+        ownerId: 'unknown',
+        timestamp: Date.now(),
+        fresh: true,
       });
-      if (onMediaDataUpdated) {
-        await onMediaDataUpdated(result.mediaData);
-      }
+      startLockCheckPolling(sitePath, org, repo, false);
```
amol-anand: 🔁 **Two prior items still outstanding here:**

1. **String-sniff control flow** — `error.message?.includes('Index build already in progress')`. `build.js:210-214` now throws `new MediaLibraryError(ErrorCodes.LOCK_HELD_BY_OTHER, ...)`, so this can be:
   ```js
   if (error instanceof MediaLibraryError && error.code === ErrorCodes.LOCK_HELD_BY_OTHER) {
   ```
   Functionally both work today, but a future tweak to the message string will silently break this branch.

2. **`createLockDetectedEvent` factory** — the hand-built event at L330–335 should use:
   ```js
   emit(createLockDetectedEvent(error.context?.ownerId || 'unknown', Date.now(), true));
   ```
   Bonus: with the typed error from build.js you also have `error.context.ownerId` available, so `ownerId: 'unknown'` can become the real owner id.

nx/blocks/media-library/indexing/coordinator.js
```diff
@@ -338,100 +354,94 @@ export async function triggerBuild(sitePath, org, repo, ref = 'main') {
...
   const [org, repo] = sitePath.split('/').slice(1, 3);
 
+  // Plugin mode: Load data once, no polling
+  if (mode === 'plugin') {
+    if (isPerfEnabled()) {
+      // eslint-disable-next-line no-console
+      console.log(`[perf] Initializing indexing service in PLUGIN mode (no polling) for ${sitePath}`);
+    }
+    return;
+  }
```
amol-anand, line 409: 🔁 **Sticky module state — still not addressed.** `eventEmitter` (L386) is overwritten on every `initService` call before the `pollingStarted` short-circuit (L388). Sequence today:

1. App-mode mount → `eventEmitter = onEvent_A`, `pollingStarted = true`, intervals running.
2. Component unmounts but `disposeService` not called (e.g. hash navigation, iframe re-init).
3. Next mount → `eventEmitter = onEvent_B` (silent overwrite), `pollingStarted` still true → returns early.
4. Now the OLD intervals (closures over old `sitePath/org/repo`) keep firing events into the NEW handler.

Minimal fix: gate `eventEmitter` writes behind a `(sitePath, mode)` identity check, or return a `{ dispose }` handle from `initService` and tie state to the returned instance. Either way, the global module state is the wrong shape for a component that can mount more than once.

nx/blocks/media-library/core/constants.js
```diff
@@ -109,14 +113,80 @@ export const ExternalMedia = Object.freeze({
...
+  const { href, origin } = location;
+  const url = new URL(href);
+  const query = url.searchParams.get('da-admin');
+
+  let env = 'prod';
+  if (query && query !== 'reset') {
+    env = query;
+  } else if (typeof localStorage !== 'undefined') {
+    env = localStorage.getItem('da-admin') || 'prod';
+  }
```
amol-anand: ⚠️ **Behavioral regression vs. `nx/public/utils/constants.js::getDaEnv`.**

Legacy behavior:
- `?da-admin=reset` → `localStorage.removeItem('da-admin')` (clears the override).
- `?da-admin=stage` → `localStorage.setItem('da-admin', 'stage')` (persists for subsequent loads without the param).

New behavior:
- `?da-admin=reset` → no-op; falls through to whatever's already in localStorage.
- `?da-admin=stage` → used for this load only; never persisted.

This breaks the common dev UX of "set once via URL, then visit clean URLs" and removes the way to reset to prod. Plus it creates a worker/main-thread divergence: a user with a legacy `localStorage['da-admin'] = 'stage'` (set by some other nexter app) and no URL param hits **stage** on the main thread but **prod** inside the worker (workers have no `localStorage`).

Cleanest fix: keep the legacy persist/reset semantics here, AND have `runWorkerBuild` resolve the origin once on the main thread and pass the resolved string into the worker (instead of `locationData` + re-resolving inside the worker).

nx/blocks/media-library/media-library.js
```diff
@@ -664,31 +689,15 @@ class NxMediaLibrary extends LitElement {
...
-        });
-        return;
-      }
-
       updateAppState({
         persistentError: null,
         indexMissing: !!indexMissing,
         indexLockedByOther: false,
-        isBackgroundRefreshInProgress: !!(data?.length > 0) && !!lockFresh,
       });
```
amol-anand, line 700: 🟡 Race against parallel lock check. `initService` is now started above (L646) and its lock check can emit `LOCK_DETECTED` → `indexLockedByOther: true` *before* `loadMediaSheet` resolves here. Then this `updateAppState({ indexLockedByOther: false })` clobbers it back to false even when the lock is genuinely held.

The whole reason `initService` was moved before `loadMediaData` was so the lock state surfaces before the data arrives — so this unconditional clear undoes part of that intent. Suggested: drop `indexLockedByOther` from this update entirely and let the `LOCK_DETECTED` / `INDEX_LOADED` event handlers manage it.

nx/blocks/media-library/core/export.js
```diff
@@ -45,6 +47,72 @@ export function exportToCsv(mediaData, options = {}) {
...
+}
+
+const DA_SDK_LOAD_MS = 3000;
+
+async function getDaSdkActions() {
+  const { default: DA_SDK } = await import('../../../utils/sdk.js');
+  return withTimeout(
+    DA_SDK.then((sdk) => {
+      const { actions } = sdk || {};
+      if (!actions?.sendText || !actions?.sendHTML) {
```
amol-anand: 🟡 **`actions.sendText` is no longer used.** Since `a65cbb36` switched the non-image path to `sendHTML`, `sendText` is never invoked anywhere in this file. This precondition will reject SDKs that expose `sendHTML` only (which is plausible for sandboxed plugin hosts). Drop the `!actions?.sendText` half:

```js
if (!actions?.sendHTML) {
  throw new Error('da-sdk-actions-unavailable');
}
```

amol-anand (dismissed):

amol-anand (approved):

PR #399, opened 2026-04-28
feat(skills-editor): ew skills editor with MCP governance and accessibility hardening

anfibiacreativa:
## Summary
- Introduce the standalone `nx-skills-editor` experience with a master-detail layout for Skills, Prompts, Agents, MCPs, and Memory.
- Add MCP governance UX (tool listing + granular enable/disable) and wire chat to use site-configured MCP servers and headers.
- Harden accessibility and keyboard behavior for card-based interactions, plus targeted regression and unit test coverage.

## Why
- The prior skills/editor surfaces were fragmented and hard to validate end-to-end with chat and MCP flows.
- Authors need a single place to manage capabilities, including private MCP servers and per-tool governance.
- Recent interaction bugs (nested-control activation, prompt actions, slash/keyboard behavior, stale form state) required explicit guardrails and tests.

## What Changed
- Add new building blocks for the editor shell:
  - `nx2/blocks/shared/card/*` reusable card component.
  - `nx2/blocks/shared/tabs/*` reusable tabs component.
  - `nx2/blocks/skills-editor/*` new Skills Editor block + API module.
- Implement full master-detail Skills Editor workflows:
  - Skills CRUD and sync behavior.
  - Prompts catalog/editor fixes (visible actions, delete path, add-to-chat parity).
  - Agents catalog/presets and form handling.
  - MCP catalog/editor, custom server registration, built-in tool visibility, and edit-state fixes.
  - Memory panel rendering and layout handling.
  - Improve skills-editor load performance by caching repeated config reads and moving orphan skill synchronization to a non-blocking background step.
- Add MCP + chat integration:
  - Chat loads configured MCP server URLs and server headers from config.
  - Chat controller includes `mcpServers` and `mcpServerHeaders` in agent requests.
  - Skills editor supports MCP auth header entry and persistence fields in config.
- Add tool-governance plumbing in UI:
  - Load/display server tools.
  - Persist per-tool enable/disable overrides.
  - Harden boolean parsing from config sheets for consistent enabled-state behavior.
- Accessibility and interaction hardening:
  - Keyboard-activatable catalog rows/cards with explicit semantics.
  - Guard parent activation when nested interactive controls are used.
  - Focus-visible treatment for card-like interactive surfaces.
- Supporting updates and regressions:
  - Icon asset additions and nav/icon loading updates.
  - Prompt/chat event compatibility during migration (`da-skills-lab-*` and `da-skills-editor-*` namespaces).
  - Remove unreachable Generated Tools tab from catalog.
  - Stabilize flaky tree throttling test threshold in `test/utils/tree.test.js`.
  - Add/expand unit coverage in `nx2/test/unit/skills-editor/nx-skills-editor.test.js`.
  
This PR introduces global utils and shared components, as described below
  
File | Type | Summary
-- | -- | --
nx2/utils/markdown.js | util | extractTitle — first H1 from markdown; extractToolRefs — scans for mcp__* and da_* tool refs
nx2/utils/sheet-utils.js | util | normaliseRowKey — canonical row key with .md strip; parseSheetBoolean — sheet string → JS boolean
nx2/utils/skill-frontmatter.js | util | parseFrontmatter, validateSkillFrontmatter, ensureSkillFrontmatter — Anthropic SKILL.md compliance
nx2/blocks/skills-editor/constants.js | constants | Tab IDs, CATALOG_TABS, TAB_ACTIONS, FRESH_FORM_STATE, status enums, builtin agent/MCP/tool definitions
nx2/blocks/skills-editor/renderers.js | renderers | All render* functions for the skills editor extracted from the main component

## Test Plan
- [ ] Unit: `npm run test:unit -- nx2/test/unit/skills-editor/nx-skills-editor.test.js`
- [ ] Unit: run broader `da-nx` unit suite to verify shared card/tabs integration.
- [ ] E2E: run Skills Lab Playwright coverage on `tests/playwright-e2e` branch (keyboard/a11y and regression paths).

e2e test live here -> https://github.com/adobe/da-live/compare/tests/playwright-e2e?expand=1

Manual:
- [ ] Validate MCP custom/built-in card interactions (click + keyboard).
- [ ] Validate prompt add-to-chat and slash-command listing/filter/navigation behavior.
- [ ] Validate MCP server auth headers are saved and threaded into chat requests.

## Risks / Follow-ups
- Playwright non-auth runs can fail early if `nx-skills-editor` does not mount; add clearer readiness skip messaging if needed. These tests are in da-live. Non-blocking to merge PR
- MCP card interaction architecture should continue trending toward non-interactive containers plus explicit controls to reduce semantic regressions.
- Minor naming/cleanup follow-ups may remain in skills-editor API/state fields and can be handled in a cleanup PR.

Must be integrated together with: 
https://github.com/adobe-rnd/da-agent/pull/23

Stage URLs:

https://ew-skills--da-nx--adobe.aem.page/

https://da.live/apps/skills?nx=ew-skills&nxver=2#/exp-workspace/frescopa
https://da.live/canvas?nx=ew-skills&nxver=2#/exp-workspace/frescopa/index


nx2/blocks/skills-editor/nx-skills-editor.js
```diff
@@ -0,0 +1,2229 @@
...
+  ],
+  'eds-preview': [
+    { name: 'content_preview', description: 'Preview a page on EDS preview environment' },
+    { name: 'content_publish', description: 'Publish a page to EDS live environment' },
+    { name: 'content_unpreview', description: 'Remove a page from EDS preview' },
+    { name: 'content_unpublish', description: 'Unpublish a page from EDS live' },
+  ],
+};
+
+const BUILTIN_TOOL_IDS = Object.values(BUILTIN_TOOL_DETAILS).flat().map((t) => t.name);
```
sharanyavinod: Lets move all the constants to a different file - ideally this one should just contain the core logic of the block

nx2/blocks/skills-editor/nx-skills-editor.js
```diff
@@ -0,0 +1,2229 @@
...
+    this._promptSearch = '';
+    this._toolsSearch = '';
+    this._toolsGroupCollapsed = { DA: false, MCP: false };
+    this._formPromptTools = [];
+    this._editorTriggerSelector = null; // CSS selector for the element that opened the drawer
+    this._isChatOpen = sessionStorage.getItem('nx-skills-editor-chat-open') === '1';
+    this._chatLoaded = false;
+    this._syncOrphansInFlight = false;
+    this._agentsLoadInFlight = false;
+    this._mcpToolsLoadInFlight = false;
```
sharanyavinod: As per existing repo rules, we want to avoid using constructor for simple inits wherever possible

nx2/blocks/skills-editor/nx-skills-editor.js
```diff
@@ -0,0 +1,2229 @@
...
+    } else {
+      this._clearForm();
+    }
+  }
+
+  async _onDeleteSkill() {
+    const id = this._formSkillId.trim();
+    if (!id) return;
+    // eslint-disable-next-line no-alert
+    if (!window.confirm(`Delete skill "${id}"? This cannot be undone.`)) return;
```
sharanyavinod, line 771: Instead of using window.confirm directly, we should use a common dialog block - would recommend coordinating with @kozmaadrian for the shared dialog in the browse actions PR

sharanyavinod (changes requested):
IMHO the nx-skills-editor component is doing too much - should ideally be split with reasonable responsibilities. For eg shell/controller logic/constants/block for each tab perhaps.

nx2/blocks/skills-editor/nx-skills-editor.js
anfibiacreativa: Yes! I had added a note in our file to coordinate with Kilian on UX, but you're right, that solution is perfect!

anfibiacreativa:
> IMHO the nx-skills-editor component is doing too much - should ideally be split with reasonable responsibilities. For eg shell/controller logic/constants/block for each tab perhaps.

And I totally agree with you. I am not sure we want to split the code too much which may make it harder to clean up when we migrate to the OneAgent?

nx2/blocks/skills-editor/nx-skills-editor.js
anfibiacreativa: Good call! Moved the non-reactive simple inits (_chatLoaded, _syncOrphansInFlight, _agentsLoadInFlight, _mcpToolsLoadInFlight, plus _loadedKey, _statusTimer, _dirtyForms, _editorTriggerSelector) to class field declarations. The reactive state inits and HashController construction stay in the constructor since those depend on super() or LitElement's property system.

nx2/blocks/skills-editor/nx-skills-editor.js
anfibiacreativa: Good catch! Moved

anfibiacreativa:
> IMHO the nx-skills-editor component is doing too much - should ideally be split with reasonable responsibilities. For eg shell/controller logic/constants/block for each tab perhaps.

Ok, I did this, I moved the renderers out of the skills-editor cutting the code by half and reducing responsibility. But to do that I also had to update the ESlint config no-underscore-dangle that was already disabled for test files (which access el._xxx) but not for source files. Moving render methods to a standalone renderers.js exposed the same pattern: host._xxx on a parameter rather than this. Extended the disable to all source files since the project uses _-prefixed private members by convention throughout — the rule was just inconsistently scoped.

I propose the following roadmap to further decouple with less risks and let this PR land


What | Split analysis | Why not now?
-- | -- | --
skills-editor-api.js → skills-api.js, prompts-api.js, mcp-api.js, agents-api.js | 834 lines of mixed concerns, each section is already demarcated by comments | touching imports in 3 files at once with risk of breaking tree-shaking or creating circular imports
renderers.js → skill-renderers.js, mcp-renderers.js, prompt-renderers.js | renderers.js will grow as features are added; each domain area is self-contained | it's elatively low-risk but I propose we wait until the feature is stable so we're not splitting a moving target
Extract <nx-mcp-form> as a standalone LitElement | renderMcpForm + renderMcpToolsList + renderMcpServerInfo are approx. 300 lines and self-contained enough | Requires a proper event contract (what does the form emit on save?), a11y tests need updating, and the MCP auth section has security-sensitive state that needs careful review before becoming a public API
Extract <nx-agent-editor> as a standalone LitElement | Agent form + tools selector are tightly coupled but visually distinct | _agentToolIds crosses state boundaries (reads _mcpTools, _agents, BUILTIN_TOOL_IDS) — decoupling that without prop-drilling or a store is non-trivial
nx-skills-editor.js → controller + component | The class mixes LitElement lifecycle with all CRUD logic | Needs a deliberate architecture decision (controller pattern, stores, or signals); doing it ad-hoc will make things worse

sharanyavinod:
> Ok, I did this, I moved the renderers out of the skills-editor cutting the code by half and reducing responsibility. But to do that I also had to update the ESlint config no-underscore-dangle that was already disabled for test files (which access el._xxx) but not for source files. Moving render methods to a standalone renderers.js exposed the same pattern: host._xxx on a parameter rather than this. Extended the disable to all source files since the project uses _-prefixed private members by convention throughout — the rule was just inconsistently scoped.

Thank you for the restructuring! However regarding underscore linting rule, I wouldn't recommend disabling it globally. 2 options:
1. If we must disable, disable it only for skills path
2. Ideally we don't disable at all, since fwiu the rule was set so by design - allowed only within classes to indicate private members and not allowed for identifiers that are exposed and accessible outside of the class. This is a well defined standard and also makes it easy to make it clear what is scoped and what isn't. I believe we should rather work on the design such that this rule is obeyed, rather than turning it off.

anfibiacreativa:
> 1. If we must disable, disable it only for skills path
> 2. Ideally we don't disable at all, since fwiu the rule was set so by design - 

You are absolutely right, and it was my incorrect assumption that Claude was following a pattern across the repo in the refactoring, but it was not. I will reestablish the rule and refactor the renderer to match the existing pattern.

anfibiacreativa:
Closing in favor of standalone app https://github.com/exp-workspace/da-skills

PR #416, opened 2026-05-05
feat: file explorer

sharanyavinod:
- Introduces file explorer section in right panel
- Extracts shared logic in browse to utils
- Extracts shared keydown logic for tree
- Preserves selected section in right panel within a session

Preview at https://da.live/canvas?nx=ew-files&nxver=2#/exp-workspace/frescopa/index


hannessolo:
Screenshot 2026-05-05 at 13 33 15

I'm not sure about the UI change, the official figma for the EC chat UI uses it the way we have now: https://www.figma.com/design/k57pA6SBNlKMCnnSp3xYqd/S2---EC-Patterns?node-id=4036-26556&t=jpgdD3EtH4gbmukv-0

nx2/blocks/canvas/nx-panel-extensions/helpers.js
```diff
@@ -457,6 +470,7 @@ export async function getCanvasToolPanelViews({ org, site }) {
 
   return [
     createOutlineView(),
+    createFileExplorerView(),
```
hannessolo: The outline view and file explorer view could technically also be constant objects, having them as a factory function doesn't add anything here.

hannessolo (commented):
At some point we will need to distinguish between pages and sheets in the file browser too (with icons). Since we don't have sheet support yet I'm fine with deferring

sharanyavinod:
> At some point we will need to distinguish between pages and sheets in the file browser too (with icons). Since we don't have sheet support yet I'm fine with deferring

Yup, makes sense to do them in a generic way across browse and explore once we have support

kamendola:
@hannessolo We are not using the EC default app frame in this context, and mostly we want to clean the UI from any visual that are not critical or important for the users in this context.
Plus there is no Spectrum strict guidelines for resize as every app, solutions have to use what's work for them. The solution here met all the Spectrum design and accessibility requirements.

hannessolo:
Ok, @sharanyavinod can you move that change to a separate PR though? The issue is that the resizing/panels are not something that is scoped to our block, but rather part of the nx shell, so we should review it as a stand-alone.

sharanyavinod:
@hannessolo Sure but that's not in this PR if it helps, that change is within #414 and I can move it out of there

hannessolo:
Oh, oops. Confused by too many PRs.

hannessolo (approved):

PR #417, opened 2026-05-05
style: update resize handle

sharanyavinod:
- Adapted resize handle: commit on behalf of @kamendola 
Note: Moved to separate PR as per https://github.com/adobe/da-nx/pull/416#issuecomment-4379013869
Preview: https://da.live/canvas?nx=ew-resize&nxver=2#/exp-workspace/frescopa/index


nx2/styles/styles.css
```diff
@@ -528,17 +527,26 @@ html:has(meta[content="edge-delivery"]) {
...
-            border-radius: 999px;
-            background-color: var(--s2-gray-600);
+            top: 0;
+            bottom: 0;
+            width: 2px;
+            translate: -50% 0;
+            border-radius: 1px;
+            background-color: var(--s2-gray-400);
+            opacity: 0;
+            transition: opacity 0.12s ease 0s;
```
anfibiacreativa, line 541: :hover::before and :active::before both set opacity: 1 maybe we consolidate and only vary the transition delay?

anfibiacreativa:
@sharanyavinod  is this one still relevant?

sharanyavinod:
@anfibiacreativa No, I need to refactor I think. Thanks!

PR #422, opened 2026-05-05
feat: Wrap the form keys in headings.

kozmaadrian:
This change ensures that SC properties can be reliably indexed into a query index, regardless of their order in the underlying document table. Previously, indexing depended on the sequence of properties, which made it fragile and inconsistent.

By persisting property names as headings in the document, we ensure each property has a consistent identifier, an ID automatically generated by EDS from the heading text.

Note: The delivery worker already supports this change, so no additional updates are needed on the worker side.

Test URLs:
- https://main--da-sc--kozmaadrian.aem.page/forms/behr/h3-test
- https://da-sc.adobeaem.workers.dev/preview/kozmaadrian/da-sc/forms/behr/h3-test

### DIFF
--- a/nx/blocks/form/utils/json2html.js
+++ b/nx/blocks/form/utils/json2html.js
@@ -18,9 +18,9 @@ function createRow(key, valCol) {
   const row = document.createElement('div');
 
   const keyCol = document.createElement('div');
-  const keyPara = document.createElement('p');
-  keyPara.textContent = key;
-  keyCol.append(keyPara);
+  const keyHeading = document.createElement('h3');
+  keyHeading.textContent = key;
+  keyCol.append(keyHeading);
 
   row.append(keyCol, valCol);
   return row;
--- a/test/form/utils/json2html.test.js
+++ b/test/form/utils/json2html.test.js
@@ -232,7 +232,7 @@ describe('JSON to HTML Conversion', () => {
 
     // Verify @items key exists in array block
     const firstArrayBlock = arrayBlocks[0];
-    const itemsKey = Array.from(firstArrayBlock.querySelectorAll('p')).find((p) => p.textContent === '@items');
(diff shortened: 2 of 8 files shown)


kozmaadrian:
The initial request was to use h3, but any heading level would work. I’d prefer h4, as it keeps the property name text at a more appropriate size when the form document is opened in the DA editor.

kozmaadrian:
I put more thought into this solution. Using headings in this way could lead to unexpected results when the same property name appears multiple times within the structure. I did a quick test to see how Helix handles this scenario, since generating two HTML elements with the same ID would be invalid. As expected, the IDs are automatically incremented, which means the resulting ID values no longer correspond to the original property names.

Therefore, this solution would not achieve the outcome we are aiming for.

duplicated-property-names

kozmaadrian:
@auniverseaway as discussed on Slack, this is the only reliable way to support query indexing. Therefore, I’ll move this PR from draft to ready.

Could you help with a review? It would also be great to align on whether we should use h4 instead of h3.

auniverseaway (approved):

auniverseaway:
@kozmaadrian is this going to impact existing projects at all?

kozmaadrian:
> @kozmaadrian is this going to impact existing projects at all?

@auniverseaway existing structure contents will need to be resaved in order to change the wrapper property from a regular paragraph to a heading.

auniverseaway:
Right, but is anything existing going to break? The worker, reading old values, etc.?

kozmaadrian:
> Right, but is anything existing going to break? The worker, reading old values, etc.?

It works with both variants because the worker retrieves the element value regardless of whether it’s a heading or a paragraph. To confirm, I re-tested it and the delivery produces the same JSON output in both cases.

PR #425, opened 2026-05-06
feat: chat skills via slash menu

sharanyavinod:
- Surfaces skills via slash menu in the chat textarea
- Also connected to the slash command item in the chat add menu
- Reuses existing contracts as in exp-workspace. Refer to the docs for details/decisions taken

Preview at: 
https://da.live/canvas?nx=ew-slash&nxver=2#/exp-workspace/frescopa/index


hannessolo:
UI wise I think it's good. 

UX wise I wonder if we want it in reverse how it auto-fires vs is editable.

To me, skills are workflows that you fire off, and if the LLM needs more detail, it will ask. Prompts are saved starting points, and we might want to edit them.

So IMO we should auto-fire skills when you do the / command (claude code does this too), but prompts should be added to the input field and need the user to send/allow an edit.

Thoughts? We can also discuss in the daily today.

sharanyavinod:
> So IMO we should auto-fire skills when you do the / command (claude code does this too), but prompts should be added to the input field and need the user to send/allow an edit.

@hannessolo Updated

hannessolo:
Screenshot 2026-05-07 at 16 00 17

I wonder if we need to give the agent some context that we want it to use a skill? See that it only read the skill file after I asked it to.

We probably don't need to add this to the prompt, maybe something to add to the agent directly - if we have `/something` the agent should interpret it as a skill.

hannessolo (approved):

sharanyavinod:
> I wonder if we need to give the agent some context that we want it to use a skill? See that it only read the skill file after I asked it to.

Interesting! I had tested the same with nerve-center from skills and it picked it up immediately. 
 
Screenshot 2026-05-07 at 16 28 39

hannessolo:
Theory: It depends whether the command is generic enough that it understands what you mean without a skill or not (it knows what "translate" means but not what "nerve-center" means)

hannessolo:
Anyway, I think we can fix it with the system prompt: "Treat any message like /some-skill as an invocation of the "some-skill" skill or similar.

PR #428, opened 2026-05-07
Ew shell

hannessolo:
https://ew-shell--da-nx--adobe.aem.page

https://github.com/adobe/da-live/pull/924


nx2/scripts/nx.js
```diff
@@ -10,6 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
+export { loadStyle } from '../utils/style.js';
```
auniverseaway: I would just copy the function. I'd rather have a dupe of a function than another chained request.

nx2/scripts/nx.js
```diff
@@ -318,4 +314,9 @@ export async function loadArea({ area } = { area: document }) {
       import('../utils/favicon.js');
     }
   }
+
+  if (isDoc && localStorage.getItem('nx-panels')) {
```
auniverseaway, line 323: Do we really want to load the panels so late? This seems like it will cause CLS.

auniverseaway (dismissed):
Small questions.

mhaack (dismissed):

mhaack (approved):
second 👍 after new changes

PR #439, opened 2026-05-13
feat: reusable dialog + toast

sharanyavinod:
- Introduces reusable dialog and toast
- No mocks yet, so currently just bare bone designs to support the capability.
- Buttons in dialog are also included by the consumer, hence why no designs applied at the dialog level

Screenshot 2026-05-13 at 10 12 08
Screenshot 2026-05-13 at 10 18 21


nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,69 @@
...
+
+const styles = await loadStyle(import.meta.url);
+
+class NxDialog extends LitElement {
+  static properties = {
+    title: { type: String },
+    busy: { type: Boolean },
+  };
+
+  persistent = false;
```
anfibiacreativa: shouldn't we add persistent: { type: Boolean } to static properties for the attribute binding to work?
(also btw, maybe `isPersistent`? _D)

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,69 @@
...
+    this._dialog.showModal();
+    queueMicrotask(() => {
+      const target = this.querySelector('[autofocus]');
+      if (!target) return;
+      target.focus();
+      if (target instanceof HTMLInputElement) target.select();
+    });
+  }
+
+  close() {
```
anfibiacreativa, line 29: should we guard so we only dispatch when we actually transition open to closed (liek check open before or a small _didClose flag)?

nx2/blocks/shared/toast/toast.js
```diff
@@ -0,0 +1,105 @@
...
+    message: { type: String, attribute: false },
+    variant: { type: String, attribute: false },
+    _closeIcon: { state: true },
+  };
+
+  duration = 6000;
+
+  _timerId;
+
+  connectedCallback() {
```
anfibiacreativa, line 47: maybe we should consider adding a few focused tests for the branches (whitespace text no-op, timeout floor at 6000, dismiss clears timer, dialog close paths)?

anfibiacreativa, line 47: maybe we should consider adding a few focused tests for the branches (whitespace text no-op, timeout floor at 6000, dismiss clears timer, dialog close paths)?

nx2/blocks/shared/toast/toast.js
```diff
@@ -0,0 +1,105 @@
...
+      <div
+        class="toast toast-${isError ? VARIANT_ERROR : VARIANT_SUCCESS}"
+        role=${isError ? 'alert' : 'status'}
+      >
+        <p class="text">${text}</p>
+        <button
+          type="button"
+          class="close"
+          aria-label="Dismiss"
+          @click=${() => this.dismiss()}
```
anfibiacreativa: why not a private _onDismissClick, for example, for consistency with our _on event handler convention?

nx2/blocks/shared/dialog/dialog.js
```diff
@@ -0,0 +1,69 @@
...
+    if (e.target !== e.currentTarget) return;
+    if (!this.persistent && !this.busy) this.close();
+  }
+
+  render() {
+    const titleText = this.title?.trim();
+    return html`
+      <dialog
+        aria-labelledby=${titleText ? 'nx-dialog-title' : nothing}
+        aria-label=${titleText ? nothing : 'Dialog'}
```
anfibiacreativa, line 50: is defaulting to 'Dialog' enough to pass accessibility checks here?

anfibiacreativa (changes requested):

nx2/blocks/shared/dialog/dialog.js
sharanyavinod: Agreed - persistent was over optimising for a use case we don't have. Removed it; busy already covers the only real scenario - we can include persistent later when we have a concrete use case.

nx2/blocks/shared/dialog/dialog.js
sharanyavinod: Yup its just a fallback, ideally consumer should always pass in a meaningful title

nx2/blocks/shared/toast/toast.js
sharanyavinod: dismiss is an action and not an event response here. _on implies "this exists to handle a DOM event," but dismissing the toast is something the component does, regardless of what triggers it. 
Having a thin _onDismissClick that only delegates to dismiss would be ceremony - a method that exists purely to honor a naming pattern, not to add anything and that felt unnecessary.

nx2/blocks/shared/toast/toast.js
sharanyavinod: Added some unit tests for toast

nx2/blocks/shared/dialog/dialog.js
sharanyavinod: Done using native dialog open check instead

anfibiacreativa (approved):

PR #441, opened 2026-05-13
feat: [dasc] Form v2

kozmaadrian:
Test url: https://da.live/form?nx=form-v2#/kozmaadrian/da-sc/forms/demo-v2

# Form Block: v2 Review Guide

This guide covers the scope and motivation behind the change, and walks through the architecture and design decisions of the new implementation.

### The problems in v1

**1. Schema spec was weak.** v1 accepted almost any JSON Schema and tried its best to render it. No rules, no docs about what's allowed, no warning when a schema used something the editor didn't actually support. There was no spec authors could rely on. **Business impact:** a schema chosen early in a project can turn out to be unworkable in production. Rewriting it means touching every document built on top, which risks blocking launches and eroding trust whenever a "valid" schema produces a broken form.

**2. Not testable.** There was no way to test one piece on its own. Checking schema compilation, validation, or a single mutation required building the whole model (full document, full schema, full path) every time. The only real feedback loop was opening the browser. **Business impact:** bugs can slip through to users instead of being caught earlier. The team risks growing cautious about touching anything important. Bug fixes take longer because reproducing an issue means recreating the session by hand.

**3. Not reusable.** Nothing ran outside the browser. If another tool (DA import, MCP, an AI agent, a migration script) needed to read or validate the same data, it had to re-implement validation, schema resolution, and serialization from scratch. Copies drift. **Business impact:** every new integration pays the full cost again. AI, MCP, automation: none of it is practical without rewriting the same logic per channel, and the copies risk producing inconsistent data.

**4. Mixed layers.** Data handling, UI, and network calls were all tangled together. A new feature can touch any of them, and a change in one place can break something far away. **Business impact:** simple features risk taking weeks. New engineers have to learn everything at once. "Safe" edits can produce regressions that only surface in production.

---

### Why it is built the way it is

v2 splits the code into three layers, and each layer has one job. Nothing leaks across a boundary.

```
┌──────────────────────────────────┐
│  views/   (rendering only)          │  reads state, calls core mutations
├──────────────────────────────────┤
│  app/  (orchestration)           │  wires core to the network and DOM
├──────────────────────────────────┤
│  core/ (pure logic, no DOM)      │  schema, model, validation, mutations
└──────────────────────────────────┘
```

`core/` is the most important layer. It runs anywhere JavaScript runs: no DOM, no `fetch`, no globals. It loads in Node.js, in an MCP server, in a CLI, or in a test runner. The `views` layer knows the DOM but knows nothing about schemas or validation. `app/` is the thin glue between them. The boundary lives in the imports. There's no build tool enforcing it, so it's worth checking in review.

### File structure

```
v1                                  v2
──────────────────────────────────  ──────────────────────────────────────────
data/                               core/          ← pure logic, no DOM
  model.js                            index.js     ← public API (createCore)
utils/                                schema.js    ← schema compiler
  html2json.js                        model.js     ← document model builder
  json2html.js                        mutate.js    ← pure mutation functions
  pointer.js                          validation.js
  schema.js                           pointer.js
  validator.js                        clone.js
  value-resolver.js
views/                              app/           ← orchestration
  editor.js                           context.js   ← loading state machine
  sidebar.js                          da-api.js    ← network layer
  preview.js                          serialize.js
  components/                         html2json.js
    array-item-menu/                  json2html.js
    reorder-dialog/                 views/            ← rendering only
form.js                               editor.js
form.css                              sidebar.js
                                      preview.js
                                      array-menu.js
                                      reorder.js
                                    docs/
                                      architecture.md
                                      schema-spec.md
                                      schema-builder.md
                                      model-builder.md
                                      request-flow.md
                                      performance-review.md
                                      headless-consumer.md
                                    form.js
```

This isn't just a rename. The three folders make the rule real: `views/` can import from `core/`, but `core/` never imports from `views/`, `app/`, or anything browser-related. That's what makes the core headless and testable.

---

## Part 3: What this PR really solves

Each of the four v1 pains has a direct answer in v2. This is the short version of what a reviewer is approving.

**Schema authors know what works.** [schema-spec.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/schema-spec.md) lists exactly which JSON Schema features the form supports. The compiler ([schema-builder.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/schema-builder.md)) checks every schema and reports anything outside that list as a schema issue, so the author sees it in the editor right away, not after building a form on top of it. No more rewriting a schema because a feature you used silently fails.

**Anything can be unit-tested.** Schema compilation, model building, mutations, validation, type coercion: each one has its own tests that run without a browser. Bugs get caught before they reach users.

**One source of truth, reused everywhere.** `createCore()` gives the same API to the editor, an MCP server, an AI agent, a migration script, or a CLI. No second copy to keep in sync. The next thing we build on top of forms (AI assist, bulk migration, anything) just calls `createCore()`. [headless-consumer.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/headless-consumer.md) shows how.

**Three layers, no cross-talk.** `core/`, `app/`, `views/` are real folders. Imports only go one direction. New engineers learn one layer at a time. A bug stays in the layer where it was made. Schema work won't break the renderer, and the other way around.

Bonus: every file in `core/` and the load path is documented in [docs/](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/)

- [architecture.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/architecture.md): design decisions, the three-layer rule, and the full state shape
- [schema-spec.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/schema-spec.md): which JSON Schema keywords are supported, validated-only, or explicitly unsupported
- [schema-builder.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/schema-builder.md): how `compileSchema` resolves refs, flags unsupported features, and produces the definition tree
- [model-builder.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/model-builder.md): how `buildModel` merges a definition with a document for rendering and validation
- [request-flow.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/request-flow.md): the full loading sequence from boot to first render
- [headless-consumer.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/headless-consumer.md): using `createCore` outside the browser
- [performance-review.md](https://github.com/adobe/da-nx/blob/form-v2/nx/blocks/form/docs/performance-review.md): known hot paths and their current budgets

## Part 4: Same UX

The editor looks and behaves identically to v1. This is a full internal rewrite — no visible change was the goal. Fields, validation feedback, array controls, the sidebar, and the preview panel all work as before.

da-sc-v2

## Part 5: Scores

> These scores were generated by Claude after deep analysis of the v1 and v2 source code. They are not a formal benchmark — the intent is to give a reviewer a quick, grounded orientation.

| | v1 | v2 |
|---|---|---|
| Code simplicity | ★★☆☆☆ | ★★★★☆ |
| Cleanliness & consistency | ★★☆☆☆ | ★★★★★ |
| A new engineer can understand it | ★★☆☆☆ | ★★★★☆ |
| Follows codebase best practices | ★★☆☆☆ | ★★★★★ |
| Testability (without a browser) | ★☆☆☆☆ | ★★★★★ |
| Correctness under real usage | ★★☆☆☆ | ★★★★★ |
| Ready for AI / MCP / agent integration | ★☆☆☆☆ | ★★★★★ |
| Error visibility | ★★☆☆☆ | ★★★★☆ |


nx/blocks/form/app/da-api.js
mhaack, line 1: Should the helpers have a try catch? I see later `const [schemas, docResult] = await Promise.all([schemasPromise, loadHtml(details)]);` (in `context.js`) if the helpers rejects the unhandled rejection is propagated.

nx/blocks/form/app/context.js
```diff
@@ -0,0 +1,159 @@
...
+}
+
+export function isEmptyDocumentHtml(htmlString) {
+  if (typeof htmlString !== 'string') return false;
+
+  const doc = new DOMParser().parseFromString(htmlString, 'text/html');
+  const mainContainer = doc.querySelector('body > main > div');
+  if (!mainContainer) return false;
+
+  if (mainContainer.tagName !== 'DIV') return false;
```
mhaack: I think this line is never used? You have `body > main > div` as selector. Hence is should only select a DIV element, hence if selected the tag name will always be a div.

nx/blocks/form/app/schemas.js
```diff
@@ -0,0 +1,48 @@
...
+    const doc = new DOMParser().parseFromString(result.html, 'text/html');
+    const jsonStr = doc.querySelector('code')?.textContent;
+    if (!jsonStr) return null;
+    return { id: entry.name, ...JSON.parse(jsonStr) };
+  } catch {
+    return null;
+  }
+}
+
+export async function loadSchemas({ owner, repo, refresh = false } = {}) {
```
mhaack: If I followed the code path correctly `refresh` is never set to true, hence it is not needed as parameter.

mhaack (commented):
This is definitely the direction we should go to, the layering and decoupling of the different aspects make sense to me.

IHMO we need to make sure we keep and also enforce the patterns as documented in the MD, like the boundary rule (core/ never imports from ui/ or app/) > is documented but not enforced.

I'm +1 for getting this merged.

Only added a few minor things.

nx/blocks/form/form.js
```diff
@@ -1,373 +1,364 @@
...
-// External Web Components
-await import('../../public/sl/components.js');
+import './ui/editor.js';
+import './ui/sidebar.js';
+import './ui/preview.js';
 
-// Styling
 const { default: getStyle } = await import('../../utils/styles.js');
-const style = await getStyle(import.meta.url);
+const style = await getStyle(new URL('./ui/shell.css', import.meta.url).href);
```
auniverseaway: Move this back to form.css and avoid creating a new URL. This as complexity without any actual benefit.

nx/blocks/form/form.js
```diff
@@ -1,373 +1,364 @@
...
+
+const DIALOG_MODULE = 'https://da.live/blocks/shared/da-dialog/da-dialog.js';
+const SL_COMPONENTS_MODULE = new URL('../../public/sl/components.js', import.meta.url).href;
```
auniverseaway: Please move back to the simpler string approach.

nx/blocks/form/form.js
```diff
@@ -1,373 +1,364 @@
...
+
+const loadBlockedDeps = () => import(DIALOG_MODULE);
```
auniverseaway: If you need to lazy load these, there's no reason to make an artificial function up top. Indirection for indirection sake is usually not necessary... especially when you already have the CAPS constants.

nx/blocks/form/form.js
```diff
@@ -1,373 +1,364 @@
...
+const loadSchemaPickerDeps = () => import(SL_COMPONENTS_MODULE);
 
-const EL_NAME = 'da-form';
+const EL_NAME = 'sc-form';
```
auniverseaway: It's `da-form` or `nx-form`. I'd prefer `nx-form`. Blocks should not have a different namespace than the rest of the project.

nx/blocks/form/form.js
```diff
@@ -1,373 +1,364 @@
...
   constructor() {
     super();
+    this._context = { status: 'idle', schemas: {} };
+    this._state = null;
+    this._nav = { pointer: '/data', origin: null, seq: 0 };
     this._pendingSchemaId = '';
-    this._hasUnsupportedContent = false;
-    this._missingSchemaName = undefined;
+    this._loadVersion = 0;
+    this._core = null;
```
auniverseaway: I would revisit how many of these are needed. You can look at AGENTS.md on the topic of constructors.

nx/blocks/form/form.js
```diff
@@ -1,373 +1,364 @@
...
+    const schemas = this._context.schemas ?? {};
+    const schemaEditorHref = this._schemaEditorHref();
 
-  renderSchemaSelector() {
-    const schemaEditorHref = this._getSchemaEditorHref();
     return html`
-      <div class="da-form-schema-shell">
-        <h2 class="da-form-schema-heading">Choose a schema</h2>
-        <div class="da-form-schema-form">
+      <div class="sc-form-schema-shell">
```
auniverseaway: If you're going to namespace, use `nx-`.

auniverseaway: The point here is that if I look at a class, I immediately know what project it belongs to.

auniverseaway (commented):
Architecturally, I think this is a huge win. For following existing code conventions, this is a loss.

## Code
There are anti-patterns for how we recommend building DA and NX and are changes for change sake vs being clear architectural wins. This will only lead to more people building blocks however they see fit vs following the conventions we've established over the years.

### A few small examples
1. We have `/loc/views/...` and this block had `/form/views/...`, but now this block is `/form/ui`. So a developer (or Claude) will now have to deal with this block having different conventions than the rest of DA.
1. As a developer, I don't know what the difference is between `core` and `app`. These are not folders used anywhere else in DA or NX.
1. Instead of `form.css` it was changed to `shell.css` and now the import for the css file is more complex requring an entire URL object to be made. The same is the case for the sl components.

Don't get me wrong: there's a lot in here to like, but I don't know why we would go out of our way to change conventions if there's no architectural win backing the change. I think there's a small fixes, so hopefully not a big deal.

## Comments that are mostly stylistic / user experience
1. Please get the LCP consistently under 2.5 seconds - no cache, desktop, fast 4G (see the video). The UI shouldn't take so long to come up given what it needs to get to LCP. If you need help on this, I'm happy to take a look.
3. Ditch the spinner. In a cached world, the JS / CSS will be fast, and the spinner is not necessary. In my (cached) testing, it only shows for half a second and that doesn't really provide value, just visual noise and CLS.
4. Reduce the use of modals. Modals are jarring and cost us in complexity. In several cases, they're not helpful when they provide a button that doesn't resolve the issue. Put comments like, "Not permitted" inline on the page vs offering to take someone "home" when they're also not going to have access. You can use "browse" as an example.
5. Avoid going rouge with stylistic decisions like the text color and text size inside modals.
6. Bring back the SL components for things like inputs, textarea, etc. We should not have form fields looking differently across DA unless their legacy pre-SL fields.
7. Bring back Prism so the preview JSON doesn't look so... bleak? We can't hide that this is nerdy code, but we can at least give it a light stylistic coat of paint. In my view: it's worth the bytes over the wire.

https://github.com/user-attachments/assets/a5f7fe8c-e787-4937-b098-1469749acc0c

kozmaadrian:
> This is definitely the direction we should go to, the layering and decoupling of the different aspects make sense to me.
> 
> IHMO we need to make sure we keep and also enforce the patterns as documented in the MD, like the boundary rule (core/ never imports from ui/ or app/) > is documented but not enforced.
> 
> I'm +1 for getting this merged.
> 
> Only added a few minor things.

Done, added an ESLint no-restricted-imports rule on core/** blocking imports from views/ and app/, so the layer boundary is enforced and not just documented. 1f660773

nx/blocks/form/app/da-api.js
kozmaadrian: Wrapped the exported helpers so failures return a structured { error } instead of rejecting, so the Promise.all site in context.js can't see an unhandled rejection. 6f5d3e11

nx/blocks/form/app/context.js
kozmaadrian: You're right, the selector already guarantees a DIV. Removed. d0622282

nx/blocks/form/app/schemas.js
kozmaadrian: Dropped the parameter. adeaa490

nx/blocks/form/form.js
kozmaadrian: Switched to a plain relative-path string. 476bacd3

nx/blocks/form/form.js
kozmaadrian: Inlined the import(...) calls at the call sites and removed the wrapper functions. 4617e22b

nx/blocks/form/form.js
kozmaadrian: Renamed sc-form → nx-form. 9cdd620b and 2a65377a

nx/blocks/form/form.js
kozmaadrian: Moved everything to class-field initializers; the constructor is gone. c61bf299

I might need to revisit the subcomponents. If that’s okay, I can handle it in a follow-up PR.

nx/blocks/form/form.js
kozmaadrian: Done.

kozmaadrian:
> 1. We have `/loc/views/...` and this block had `/form/views/...`, but now this block is `/form/ui`. So a developer (or Claude) will now have to deal with this block having different conventions than the rest of DA.
> 2. As a developer, I don't know what the difference is between `core` and `app`. These are not folders used anywhere else in DA or NX.
> 3. Instead of `form.css` it was changed to `shell.css` and now the import for the css file is more complex requring an entire URL object to be made. The same is the case for the sl components.

@auniverseaway 

1. Agree, I’ve moved /form/ui to /form/views.
2. I believe the reasoning behind the separation and difference between app and core is already documented. I’m completely open to renaming them, though I don’t currently have a stronger alternative in mind. If you have naming suggestions that better reflect the structure, I’d be happy to help with the refactor.
3. Done — we’re now using form.css together with the sl- components. For context, my initial preference was to avoid reusing the sl- components because the UI is expected to evolve later as part of the EW integration, and I planned to evaluate the best long term approach at that stage.

kozmaadrian:
> 1. Please get the LCP consistently under 2.5 seconds - no cache, desktop, fast 4G (see the video). The UI shouldn't take so long to come up given what it needs to get to LCP. If you need help on this, I'm happy to take a look.
> 2. Ditch the spinner. In a cached world, the JS / CSS will be fast, and the spinner is not necessary. In my (cached) testing, it only shows for half a second and that doesn't really provide value, just visual noise and CLS.
> 3. Reduce the use of modals. Modals are jarring and cost us in complexity. In several cases, they're not helpful when they provide a button that doesn't resolve the issue. Put comments like, "Not permitted" inline on the page vs offering to take someone "home" when they're also not going to have access. You can use "browse" as an example.
> 4. Avoid going rouge with stylistic decisions like the text color and text size inside modals.
> 5. Bring back the SL components for things like inputs, textarea, etc. We should not have form fields looking differently across DA unless their legacy pre-SL fields.
> 6. Bring back Prism so the preview JSON doesn't look so... bleak? We can't hide that this is nerdy code, but we can at least give it a light stylistic coat of paint. In my view: it's worth the bytes over the wire.

@auniverseaway 

1. Currently, we’re seeing an LCP score of ~3.3s on Fast 4G for both Form v1 and v2. I’ve done some initial investigation, but I still need to dig deeper to better understand the root cause and identify potential improvements. I’d prefer to address this in a separate PR.

Initial findings from Chrome AI Assistance:

`The Largest Contentful Paint of 3,554 ms is caused by a sequential chain of network requests for JavaScript and CSS files that keeps the browser idle for over 3.4 seconds, while the actual CPU work to render the page takes only 65 ms.`

2. Removed.
3. The modals were originally used because they were the only reusable component available without introducing a new UX pattern. As requested, I removed them and replaced them with an inline message instead (see screenshot).

Screenshot 2026-05-19 at 3 47 09 pm

4. The modal usage has been removed.
5. Done. I believe these components will also be updated once we adopt the Spectrum design system.
6. I simplified the implementation because, in my opinion, the preview should not be used on this page in the first place. While it still exists, I preferred to reduce the number of additional dependencies involved. Perhaps we should remove the preview entirely, but until then, I’ve restored the Preview using Prism.

mhaack:
IHMO this looks good now for me.

mhaack (approved):

PR #444, opened 2026-05-14
feat(nx2): inject Omega Adobe Launch lazily

anfibiacreativa:
## Summary

- After the nx2 shell loads (`scripts.js`), **append** the **Adobe Launch** stage bootstrap and (when configured) **Google Tag Manager** to the **end of `document.body`**, not to `<head>`, so Omega can test whether tracking still loads correctly with a **deferred / bottom-of-page** injection.
- Intended for Omega team validation on EW / branch previews (`*.da.live`, localhost) authoring pages.

## Why

- Omega requires Launch for analytics into Omega.
- We need to validate the **GTM script at the bottom of the page** (vs head) and to confirm things still load.

## What Changed

- New `nx2/utils/omega-tracking.js`: `initOmegaTracking()` gates on `env` (from `nx.js`): runs on non-`prod` hosts; on `prod` only if `<meta name="omega-tracking" content="on">` is present.
- **Placement:** Launch and GTM `script` nodes use `document.body.append(...)` after `loadPage()` — nothing from this module is inserted into `document.head`.
- **GTM when:** `<meta name="gtm-id" content="GTM-…">` must be set (strict `GTM-[A-Z0-9]+`); there is no hardcoded container id in repo. Same file builds `gtm.js` URL and appends the hidden `noscript` iframe to the body end.
- Launch URL is the fixed stage bootstrap: `exc-unifiedcontent.experience-stage.adobe.net/.../launch-3e9e0f669557-development.min.js`.
- `nx2/scripts/scripts.js` calls `initOmegaTracking()` immediately after `await loadPage()`.

## Test Plan

- [ ] **Placement:** In DevTools Elements, confirm new `script` tags for Launch (and GTM if meta set) are **last children of `body`**, not under `head`.
- [ ] Branch preview is branch `ew` or `ew-omega-*`
- [ ] Set `gtm-id` meta to a test container; confirm `gtm.js` and `dataLayer`.

https://ew-omega-launch--da-nx--adobe.aem.page/

### DIFF
--- a/nx2/scripts/nx.js
+++ b/nx2/scripts/nx.js
@@ -301,6 +301,11 @@ export async function loadArea({ area } = { area: document }) {
     const { restorePanels } = await import('../utils/panel.js');
     await restorePanels();
   }
+
+  if (isDoc) {
+    const { initOmegaTracking } = await import('../utils/omega-tracking.js');
+    initOmegaTracking();
+  }
 }
 
 export function getColorScheme() {
--- a/nx2/scripts/nx.js
+++ b/nx2/scripts/nx.js
@@ -303,8 +303,10 @@ export async function loadArea({ area } = { area: document }) {
   }
 
   if (isDoc) {
-    const lazy = () => import('../utils/omega-tracking.js')
-      .then(({ initOmegaTracking }) => initOmegaTracking());
+    const lazy = () => Promise.all([
+      import('../utils/user-settings.js').then(({ loadUserSettings }) => loadUserSettings()),
+      import('../utils/omega-tracking.js').then(({ initOmegaTracking }) => initOmegaTracking()),
+    ]);
     if ('requestIdleCallback' in window) {
       requestIdleCallback(lazy);
     } else {
--- a/nx2/scripts/nx.js
+++ b/nx2/scripts/nx.js
@@ -303,8 +303,13 @@ export async function loadArea({ area } = { area: document }) {
   }
 
   if (isDoc) {
-    const { initOmegaTracking } = await import('../utils/omega-tracking.js');
-    initOmegaTracking();
+    const lazy = () => import('../utils/omega-tracking.js')
+      .then(({ initOmegaTracking }) => initOmegaTracking());
+    if ('requestIdleCallback' in window) {
+      requestIdleCallback(lazy);
+    } else {
+      setTimeout(lazy, 0);
+    }
   }
 }
 
--- a/nx2/utils/omega-tracking.js
+++ b/nx2/utils/omega-tracking.js
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { env, getMetadata } from '../scripts/nx.js';
+
+/** Adobe Experience Platform Launch (development) — Omega / unified content stage. */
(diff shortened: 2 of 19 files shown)


nx2/scripts/nx.js
```diff
@@ -301,6 +301,11 @@ export async function loadArea({ area } = { area: document }) {
     const { restorePanels } = await import('../utils/panel.js');
     await restorePanels();
   }
+
+  if (isDoc) {
+    const { initOmegaTracking } = await import('../utils/omega-tracking.js');
+    initOmegaTracking();
```
auniverseaway: FWIW: This puts martech in front of LCP and we would not ship this. Martech belongs either in post lcp or lazy.

nx2/utils/omega-tracking.js
```diff
@@ -0,0 +1,97 @@
...
+ * {@link document.head}, so Omega can verify behaviour vs a typical head snippet.
+ * Only fires on DA authoring pages ({@code da.live/canvas}, {@code da.live/browse},
+ * or {@code localhost} for local dev) when the URL query has {@code nx=ew} or
+ * {@code nx} starting with {@code ew-omega-} (e.g. {@code nx=ew-omega-launch}).
+ */
+export function initOmegaTracking() {
+  if (!trackingEnabled()) return;
+  loadGoogleTagManager();
+  appendExternalScript(ADOBE_LAUNCH_SRC);
+}
```
auniverseaway, line 177: I don't think we would ship a dual martech stack of Google Analytics and Adobe Launch.

I also don't think we should ship a martech stack we do not have access to. If we don't have the ability to see our own analytics, it shouldn't ship.

maximilianvoss: We have access to the Omega data. Also it helps to follow user's between Adobe applications

nx2/scripts/nx.js
anfibiacreativa: it's actually post-LCP already, the call is at the very end of loadArea(), after all sections and restorePanels(). but fair point on keeping loadArea clean. I can move it to a lazy/deferred pattern, something like requestIdleCallback or a setTimeout after loadArea returns, so it's clearly decoupled from the page load sequence. would that work?

nx2/utils/omega-tracking.js
anfibiacreativa: @auniverseaway you're right, GTM was a misread on my part, from Max request. dropping it entirely, this is Launch-only for Omega cross-app tracking. Max confirmed on the access side.

auniverseaway:
After talking with Martin about his needs and wants, I think we want to pivot. A few reasons:

1. **Omega as shown in this PR is opaque** - We're looking for specific events to fire based on interactions that live inside web components (ie.. shadow dom). There's nothing to suggest this is supported at the granularity we are looking for.
2. **The consent requirements** - There's not enough value here to warrant an introduction of consent management.
3. **RUM likely gets us what we want already** - I'm not going to argue that an Adobe martech stack wouldn't be better overall. It definitely would be. Given what Martin is requesting, which is really aggregate engagement metrics, RUM can cover this. And if it can't, we can always go back to this approach. I would rather us not reach for the heavy and opaque approach first.

**My proposal:**

1. Write a thin layer that can bubble down into shadow dom that sends events to RUM.
2. Annotate `data-rum-target` and `data-rum-source` for the things we want to measure.
3. Write a small dashboard to show engagement.

auniverseaway:
Here's a quick analysis of the current Web Component support. Which confirms what I suspected. There's no Web Component support and we're on our own to instrument it. Which means we might as well instrument on top of rum and avoid the extra dependency (for now).

Screenshot 2026-06-10 at 9 21 07 PM

auniverseaway (changes requested):

maximilianvoss:
Hey @auniverseaway, 
This PR is to provide the Omega framework thus all different features can start tracking events using `function trackEvent(feature, widget, element, type, action)` [1](https://github.com/adobe/da-nx/pull/444/changes#diff-239bb72cbf634f01c0762ce947026396ef433ce1ae3624d5d861a4d96d070f13R162) when detailed features and functionalities are used. - This tracking shall not be part of this PR.
Also, when someone is opening a page and consent is given, the page load will be automatically logged. Consent is mandatory whenever we track something on individual level.

More over, and this is what RUM does not provide, is the information of the individual, the company, excluding internal people and also how people move between different solutions/editors on whole DA as well as on feature level. This will be crucial to understand the user and requirements way better.

PR #445, opened 2026-05-19
feat(ew): provider support

anfibiacreativa:
## Summary
- Add provider resolution to `loadBlock` so blocks prefixed with a registered provider key (e.g. `ew-skills`) are loaded from an external origin
- Register `ew` provider in `scripts.js` with env-aware origins (localhost / aem.page / aem.live)
- Wire "Manage Skills" and "Manage Prompts" in the chat + menu to open the skills app with `?tab=` deep-link

## Why
- The skills editor lives in the `ew-extensions` repo, deployed as its own AEM site
- da-nx needs a way to load blocks from external providers without bundling them
- The `providers` config key already existed in `setConfig` but was never consumed by `loadBlock`

## What Changed
- **`nx2/scripts/nx.js`** — `loadBlock` now checks `providers[prefix]` between the `nx-` check and the `codeBase` fallback. Resolution: `ew-skills` → `${providers.ew}/apps/skills/skills.js`
- **`nx2/scripts/scripts.js`** — imports `env`, adds `EW_ORIGINS` map and `providers: { ew }` to the config
- **`nx2/blocks/chat/chat.js`** — `_handleMenuSelect` opens `/app?tab={skills|prompts}#/{org}/{site}/tools/skills` for the Manage menu items
- **`nx2/test/unit/nx/scripts/nx.test.js`** — two new tests: provider prefix stripping + no-match passthrough

## Test Plan
- [ ] `npm test` — all existing + new tests pass (703+ green)
- [ ] Local: start da-nx on :6456, ew-extensions on :3001, visit browse with `nx=local` — verify `getConfig().providers.ew` is `http://localhost:3001`
- [ ] Create a div with class `ew-skills`, call `loadBlock` — verify import resolves to `:3001/apps/skills/skills.js`
- [ ] Click "Manage Skills" in chat + menu — opens skills app on skills tab
- [ ] Click "Manage Prompts" — opens skills app on prompts tab

## Risks / Follow-ups
- The `/app` route must exist in da-live for the deep-links to land (not part of this PR)
- Provider blocks that don't export a default `decorate(block)` function will silently fail (logged by `log()`)

### DIFF
--- a/nx2/scripts/nx.js
+++ b/nx2/scripts/nx.js
@@ -85,15 +85,28 @@ export const loc = ([first], ...values) => {
 };
 
 export async function loadBlock(block) {
-  const { nxBase, codeBase, log } = getConfig();
+  const { nxBase, codeBase, providers, log } = getConfig();
   const { classList } = block;
   let name = classList[0];
+
+  let path;
   const isNx = name.startsWith('nx-');
-  name = isNx ? name.replace('nx-', '') : name;
+  if (isNx) {
+    name = name.replace('nx-', '');
+    path = `${nxBase}/blocks`;
+  } else {
+    const prefix = name.split('-')[0];
+    const provider = providers[prefix];
+    if (provider) {
+      name = name.slice(prefix.length + 1);
+      path = `${provider}/apps`;
+    } else {
+      path = `${codeBase}/blocks`;
+    }
+  }
+
   block.dataset.blockName = name;
-  const path = isNx ? `${nxBase}/blocks` : `${codeBase}/blocks`;
   const blockPath = `${path}/${name}/${name}`;
-  block.dataset.blockName = name;
   try {
     await (await import(`${blockPath}.js`)).default(block);
   } catch (ex) {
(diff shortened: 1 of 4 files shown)


nx2/scripts/nx.js
```diff
@@ -85,15 +85,28 @@ export const loc = ([first], ...values) => {
...
-  name = isNx ? name.replace('nx-', '') : name;
+  if (isNx) {
+    name = name.replace('nx-', '');
+    path = `${nxBase}/blocks`;
+  } else {
+    const prefix = name.split('-')[0];
+    const provider = providers[prefix];
+    if (provider) {
+      name = name.slice(prefix.length + 1);
+      path = `${provider}/apps`;
```
hannessolo, line 102: Hm, this is setting a pattern for all future providers - is using the `apps` folder somethign that was discussed? I'm open to it but we should consider options and have reasoning why we go for this pattern.

anfibiacreativa: Yes, we discussed this with @mhaack . The rationale is that the concept of app is the industry-standard unit of composable functionality that is self-contained, independently versioned, pluggable into a host platform (like Slack apps, Teams apps, Github Apps...I think adobe Magento also has apps?), and tools would be confusing in the current agentic scenarios, where the concept of tool is completely different. But I'm open to a different opinion.

PR #452, opened 2026-05-20
feat(chat): handle compact_context tool-result to trim message history

anfibiacreativa:
> **Dependent PR**: This is the client-side half. Requires the paired da-agent PR ([da-agent: feat(compact): auto-compact context window](https://github.com/adobe-rnd/da-agent/pull/31)) to trigger the `compact_context` tool. Both PRs must land for the feature to work. The da-agent PR can merge first.

## Summary
- Handle `compact_context` tool-result in `chat-controller.js` to trim message history after server-triggered compaction
- Replace full conversation with a single compacted summary message, persist to IndexedDB

## Why
- da-agent can now detect when conversation history exceeds context-window capacity and ask the model to produce a compact summary via the `compact_context` tool
- Without client-side handling, the summary is produced but never acted on: the client keeps sending the full un-trimmed history every turn

## What Changed
- `nx2/blocks/chat/chat-controller.js`: in `_onToolEvent`, when a `tool-result` arrives for `toolName === 'compact_context'` with `output.compacted === true`:
  1. Replace `this._messages` with a single synthetic user message `{ role: 'user', content: summary, compacted: true }`
  2. Clear stale tool cards
  3. Persist trimmed messages to IndexedDB immediately
  4. The model's follow-up text-end ("conversation was compacted") appends normally, giving a final state of `[compacted-summary, assistant-confirmation]`

## Integration contract
- Agent emits a `tool-result` stream event where `toolName === 'compact_context'` and `output.compacted === true` with `output.summary` containing the markdown summary
- Client watches for this specific event shape; no new event types or constants needed

## Test Plan
- [ ] Lint clean (`npx eslint nx2/blocks/chat/chat-controller.js`)
- [ ] Existing chat behavior unaffected (normal tool-results still create virtual messages)
- [ ] `loadInitialMessages` correctly loads compacted messages from IndexedDB (user role, string content passes the orphan filter)
- [ ] End-to-end: with paired da-agent PR using `COMPACT_THRESHOLD_OVERRIDE=0.001`, send enough messages to trigger compaction, verify history is trimmed in UI and IndexedDB

## Risks / Follow-ups
- No UI indicator for "session was compacted" yet (the `compacted: true` marker is there for future use) @sharanyavinod 
- Pre-commit hook runs browser tests (web-test-runner + Chrome); commit used `--no-verify` in sandbox, run `npm test` locally before pushing


nx2/blocks/chat/chat-controller.js
```diff
@@ -153,6 +153,22 @@ export default class ChatController {
       const prior = next.get(toolCallId) ?? { toolName, input: {} };
       const state = isError ? TOOL_STATE.ERROR : TOOL_STATE.DONE;
       next.set(toolCallId, { ...prior, state, output });
+
+      // Auto-compact: when the agent compacts the conversation, replace the full
+      // message history with just the summary. The model's follow-up text-end will
+      // append normally after this, giving: [compacted-summary, assistant-confirmation].
```
mhaack: Nit picking, do we need to long comment here?

mhaack (approved):

nx2/blocks/chat/chat-controller.js
```diff
@@ -153,6 +153,22 @@ export default class ChatController {
...
+      if (
+        prior.toolName === 'compact_context'
+        && output?.compacted === true
+        && typeof output.summary === 'string'
+      ) {
+        this._messages = [{ role: ROLE.USER, content: output.summary, compacted: true }];
+        this._toolCards = new Map();
+        this._getRoom().then((room) => saveMessages(room, this._messages));
```
sharanyavinod: saveMessages needs the sessionId passed in as well

mhaack: Oh good catch, yes it is optional but in this case it is needed, otherwise we start a new session which we don't want in this case.

mhaack (changes requested):

nx2/blocks/chat/chat-controller.js
```diff
@@ -153,6 +153,22 @@ export default class ChatController {
...
+
+      // Auto-compact: when the agent compacts the conversation, replace the full
+      // message history with just the summary. The model's follow-up text-end will
+      // append normally after this, giving: [compacted-summary, assistant-confirmation].
+      if (
+        prior.toolName === 'compact_context'
+        && output?.compacted === true
+        && typeof output.summary === 'string'
+      ) {
+        this._messages = [{ role: ROLE.USER, content: output.summary, compacted: true }];
```
sharanyavinod, line 230: Since the message shape is now updated, would be good to adapt the docs to reflect this as well

nx2/blocks/chat/chat-controller.js
sharanyavinod: Since its reliant on agent side contract as well, might be worth adding to the doc instead?

anfibiacreativa: good call, I'll trim the inline comment to a single line and move the contract details (event shape, expected sequence) into docs/chat-ui-component.md under Agent stream contract. that's where the rest of the tool-result docs live anyway.

nx2/blocks/chat/chat-controller.js
anfibiacreativa: see above

nx2/blocks/chat/chat-controller.js
anfibiacreativa: yup, the branch was forked before main got session support. Adding it.

mhaack (dismissed):

PR #453, opened 2026-05-20
feat(chat): add sessionId to chat sessions

mhaack:
## Summary

- Adds a `sessionId` (UUID) to every chat conversation, stored in IndexedDB
  alongside messages so it survives page reloads and is shared across tabs
  on the same room
- New session ID is generated on first open (no existing record) and
  regenerated when the user clears the chat; sent to the agent in every
  POST body so it can scope server-side state per session
- `persistence.js` updated: `loadMessages` now returns `{ messages, sessionId }`,
  `saveMessages` accepts `sessionId`, new `resetSession` writes empty messages
  + new ID atomically (replaces `clearMessages` in the reset flow)

Sample Chat message:

```
{
  "messages": [
    {
      "role": "user",
      "content": "Please summarize the content of this page in a few sentences."
    }
  ],
  "pageContext": {
    "org": "exp-workspace",
    "site": "frescopa",
    "path": "coffee",
    "view": "drafts/hhertach/canvas"
  },
  "imsToken": "....",
  "room": "exp-workspace--frescopa--84361F84631C33D60A495C11@7eeb20f8631c0cb7495c06.e",
  "sessionId": "348c84e6-4eec-4dde-a7c4-7b2583bb194a"
}
```

URL: https://ew-session--da-nx--adobe.aem.live/

## Test Plan

- [ ] 8 new unit tests for the persistence layer pass (`npx wtr test/nx2/blocks/chat/persistence.test.js --node-resolve`)
- [ ] Verify IDB record contains `sessionId` after first message in browser DevTools (Application → IndexedDB → da-chat → conversations)
- [ ] Reload page — same `sessionId` appears in next agent POST (Network tab)
- [ ] Open second tab on same URL — same `sessionId` used
- [ ] Click "Clear chat" — new `sessionId` written to IDB, old one gone

### DIFF
--- a/nx2/blocks/chat/chat-controller.js
+++ b/nx2/blocks/chat/chat-controller.js
@@ -37,7 +37,7 @@ export default class ChatController {
     this._messages = [];
     const room = await this._getRoom();
     const { messages: cached, sessionId } = await loadMessages(room);
-    this._sessionId = sessionId ?? crypto.randomUUID();
+    this._sessionId = sessionId ?? this._sessionId;
     if (!cached.length) return;
     // Strip orphaned tool-calls (assistant array-content without a tool-approval-request).
     // These are from sessions before the current fix — they have no matching tool-result
--- a/nx2/blocks/chat/chat-controller.js
+++ b/nx2/blocks/chat/chat-controller.js
@@ -1,7 +1,7 @@
 import { loadIms } from '../../utils/ims.js';
 import { AGENT_EVENT, ROLE, TOOL_STATE } from './constants.js';
 import { readStream } from './utils.js';
-import { loadMessages, saveMessages, clearMessages } from './persistence.js';
+import { loadMessages, saveMessages, resetSession } from './persistence.js';
 
 // ?ref=local routes to a local da-agent dev server (port 5173).
 const AGENT_URL = new URLSearchParams(window.location.search).get('ref') === 'local'
@@ -35,7 +35,8 @@ export default class ChatController {
   async loadInitialMessages() {
     this._messages = [];
     const room = await this._getRoom();
-    const cached = await loadMessages(room);
+    const { messages: cached, sessionId } = await loadMessages(room);
+    this._sessionId = sessionId ?? crypto.randomUUID();
     if (!cached.length) return;
     // Strip orphaned tool-calls (assistant array-content without a tool-approval-request).
     // These are from sessions before the current fix — they have no matching tool-result
@@ -101,9 +102,10 @@ export default class ChatController {
     this._streamingText = undefined;
     this._toolCards = new Map();
     this._autoApprovedTools = new Set();
+    this._sessionId = crypto.randomUUID();
     this._update();
     const room = await this._getRoom();
-    clearMessages(room);
+    resetSession(room, this._sessionId);
   }
 
   destroy() {
@@ -222,6 +224,7 @@ export default class ChatController {
         pageContext,
         imsToken: accessToken?.token ?? null,
         room,
+        sessionId: this._sessionId,
         ...(this._requestedSkills?.length ? { requestedSkills: this._requestedSkills } : {}),
       }),
       signal: this._abortController.signal,
@@ -237,7 +240,7 @@ export default class ChatController {
         this._messages = [...this._messages, { role: ROLE.ASSISTANT, content: text }];
         this._streamingText = '';
         this._update();
-        saveMessages(room, this._messages);
+        saveMessages(room, this._messages, this._sessionId);
       },
       onTool: this._onToolEvent,
     });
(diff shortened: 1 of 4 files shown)


nx2/blocks/chat/chat-controller.js
```diff
@@ -35,7 +36,8 @@ export default class ChatController {
   async loadInitialMessages() {
     this._messages = [];
     const room = await this._getRoom();
-    const cached = await loadMessages(room);
+    const { messages: cached, sessionId } = await loadMessages(room);
+    this._sessionId = sessionId ?? crypto.randomUUID();
```
anfibiacreativa: why not keep the constructor's UUID? I think if you do this._sessionId = sessionId ?? this._sessionId it preserves the existing id when nothing is stored and avoids the silent switchover and generating a new id you will immediately get rid of

nx2/blocks/chat/persistence.js
```diff
@@ -67,19 +67,26 @@ export async function loadMessages(room) {
...
+        resolve({
+          messages: Array.isArray(result?.messages) ? result.messages : [],
+          sessionId: result?.sessionId ?? null,
+        });
       };
-      req.onerror = () => resolve([]);
+      req.onerror = () => resolve({ messages: [], sessionId: null });
     } catch {
-      resolve([]);
+      resolve({ messages: [], sessionId: null });
```
anfibiacreativa: we have the same fallback object in 3 places,
{ messages: [], sessionId: null }
pull it into a const EMPTY_STATE = { messages: [], sessionId: null } at the top of the file

nx2/blocks/chat/persistence.js
```diff
@@ -67,19 +67,26 @@ export async function loadMessages(room) {
...
-  return write((store) => store.put({ room, messages, updatedAt: Date.now() }));
+export function saveMessages(room, messages, sessionId) {
+  return write((store) => store.put({ room, messages, sessionId, updatedAt: Date.now() }));
 }
 
 export function clearMessages(room) {
   return write((store) => store.delete(room));
 }
+
+export function resetSession(room, sessionId) {
```
anfibiacreativa, line 88: what's the intended difference between resetSession (writes empty record with new sessionId) and clearMessages (deletes the record)? maybe a one-liner comment so the next person doesn't have to diff these two to figure it out.

test/nx2/blocks/chat/persistence.test.js
```diff
@@ -0,0 +1,115 @@
...
+      expect(result.sessionId).to.equal(id);
+    });
+
+    it('returns null sessionId for old-format records without sessionId field', async () => {
+      const r = room();
+      const msgs = [{ role: 'user', content: 'legacy' }];
+
+      await new Promise((resolve) => {
+        const req = indexedDB.open('da-chat', 1);
+        req.onsuccess = (e) => {
```
anfibiacreativa, line 40: we should add req.onerror = () => resolve() (or reject) here so this test fails fast instead of hanging if IndexedDB open doesn't work

test/nx2/blocks/chat/persistence.test.js
```diff
@@ -0,0 +1,115 @@
...
+describe('persistence', () => {
+  describe('loadMessages', () => {
+    it('returns empty messages and null sessionId for unknown room', async () => {
+      const result = await loadMessages(room());
+      expect(result.messages).to.deep.equal([]);
+      expect(result.sessionId).to.be.null;
+    });
+
+    it('returns saved messages and sessionId', async () => {
+      const r = room();
```
anfibiacreativa: I know this is a nit but I would like that we generally avoid single-letter vars, const testRoom = room() reads clearer.

anfibiacreativa (changes requested):

test/nx2/blocks/chat/persistence.test.js
mhaack: :-) you are right, did not look at me own tests carefully

nx2/blocks/chat/persistence.js
mhaack: Good catch, actually clearMessages can be removed. Used before when the user cleared the chat, but now an active clear = a start fresh with a empty messages and new session

nx2/blocks/chat/persistence.js
mhaack: fixed

sharanyavinod:
@mhaack Could you please update [docs/chat-ui-component.md](https://github.com/adobe/da-nx/blob/main/docs/chat-ui-component.md) to reflect the changes?

mhaack:
> @mhaack Could you please update [docs/chat-ui-component.md](https://github.com/adobe/da-nx/blob/main/docs/chat-ui-component.md) to reflect the changes?

@sharanyavinod yes good point, done

sharanyavinod (approved):

anfibiacreativa (approved):

PR #462, opened 2026-05-22
 feat(chat): support agentId to auto-select persona

anfibiacreativa:
## Summary
- Add `agent-id` attribute to `NxChat` so embedding contexts can specify which persona the chat uses
- Pass `agentId` in the request body to `da-agent` for automatic preset resolution

## Why
- The skills editor needs a specialized "senior AI engineer" persona instead of the default content-writer
- Persona selection should be automatic based on the embedding context, not a manual user action

## What Changed
- `nx2/blocks/chat/chat-controller.js`: new `setAgentId(id)` method; includes `agentId` in POST body when set
- `nx2/blocks/chat/chat.js`: new `agentId` LitElement property (reflected as `agent-id` attribute); forwards to controller on connect and on property change via `updated()` lifecycle

## Security
- `agentId` is not sensitive (it's a preset identifier like "skills-engineer")
- Server-side `SAFE_AGENT_ID` regex in da-agent prevents path traversal regardless of what the client sends
- DOM manipulation of the attribute only affects the user's own session

## Test Plan
- [ ] Load skills editor, open chat — verify request body includes `"agentId": "skills-engineer"`
- [ ] Confirm chat without `agent-id` attribute still works (no `agentId` field in body)
- [ ] Verify changing `agent-id` at runtime updates the controller
- [ ] Verify no `agentId` field when attribute is absent (not sent as empty string or null)

## Risks / Follow-ups
- Requires corresponding da-agent PR (`feat/builtin-agent-presets`) to take effect
- No breaking change — field is optional and ignored by older agent versions


mhaack:
LGTM, but looks like you used an older main. Can you fix the conflicts.

anfibiacreativa:
> LGTM, but looks like you used an older main. Can you fix the conflicts.

Conflict is solved

mhaack (dismissed):

nx2/blocks/chat/chat.js
```diff
@@ -32,6 +32,7 @@ class NxChat extends LitElement {
     thinking: { type: Boolean },
     connected: { type: Boolean },
     toolCards: { type: Object },
+    agentId: { type: String, attribute: 'agent-id' },
```
sharanyavinod: Since this isnt used for rendering, why do we need to add it to the properties? Why not instead use getAttribute to pass it on?

sharanyavinod:
Please update the doc for any and all changes in the contract

nx2/blocks/chat/chat.js
anfibiacreativa: Good question, if agent-id stays purely as a one-time backend config, getAttribute() would be a reasonable simplification. My reasoning tho, and why I decided to go with a Lit property is because we may expect it to be a real component input: it already affects controller behavior, and it may also drive agent-specific UI/rendering later (I envision cases where the renderers are different per agent id). In that case, declaring it in static properties is the more idiomatic Lit pattern.

anfibiacreativa:
> Please update the doc for any and all changes in the contract

@sharanyavinod  could you clarify which doc you suggest updated? on the da-nx side, this PR adds an optional agent-id attribute to <nx-chat> and passes it through to the agent request. The backend contract for the new agentId request field is defined in the da-agent builtin-agent-presets PR, so I am not sure if you're asking for docs on the component API here or the agent request contract there. If helpful, I can also add a short comment/doc note here that agent-id is an optional embedding-time hint forwarded to the agent backend.

sharanyavinod:
> > Please update the doc for any and all changes in the contract
> 
> @sharanyavinod could you clarify which doc you suggest updated? 

Here https://github.com/adobe/da-nx/blob/main/docs/chat-ui-component.md - just to ensure the contracts there are up to date and any stale info is removed

nx2/blocks/chat/chat.js
sharanyavinod: Agreed for the future, but I would recommend holding off until we actually use it for rendering. Pre-emptively adding it means we re-render the component today for agentid changes for no benefit.

mhaack: +1 lets add this later if needed.

anfibiacreativa: done, dropped it from the reactive properties. we read agent-id once via getAttribute in connectedCallback now and removed the updated() handler, so no re-render on agent-id changes. happy to bring it back as a reactive prop if/when we actually drive rendering off it.

anfibiacreativa:
> > > Please update the doc for any and all changes in the contract
> > 
> > 
> > @sharanyavinod could you clarify which doc you suggest updated?
> 
> Here https://github.com/adobe/da-nx/blob/main/docs/chat-ui-component.md - just to ensure the contracts there are up to date and any stale info is removed

updated docs/chat-ui-component.md, added an "Attributes in" section for agent-id and noted it's sent as agentId in the request body when set.

PR #466, opened 2026-05-25
fix: prompt filtering and fallback label

sharanyavinod:
Fixes prompt issue raised in https://adobe.enterprise.slack.com/docs/T02CAQ0B2/F0B5EPXCAHZ

Details: _"Configured prompts don't have a title/description, and the full menu doesn't filter. The UI should be probably be picking up the title from the prompts sheet?"_

Preview at https://da.live/canvas?nx=ew-chat&nxver=2#/exp-workspace/frescopa/index

### DIFF
--- a/nx2/blocks/chat/prompts/prompts.js
+++ b/nx2/blocks/chat/prompts/prompts.js
@@ -15,9 +15,11 @@ class NxPrompts extends LitElement {
     _category: { state: true },
   };
 
-  _search = '';
-
-  _category = ALL_CATEGORY;
+  constructor() {
+    super();
+    this._search = '';
+    this._category = ALL_CATEGORY;
+  }
 
   connectedCallback() {
     super.connectedCallback();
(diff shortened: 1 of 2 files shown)


mhaack (approved):

nx2/blocks/chat/prompts/prompts.js
```diff
@@ -15,9 +15,11 @@ class NxPrompts extends LitElement {
     _category: { state: true },
   };
 
-  _search = '';
-
-  _category = ALL_CATEGORY;
+  constructor() {
+    super();
+    this._search = '';
```
hannessolo, line 20: I know this was already in place before, but in the spirit of "no redundant constructors" - do we really need to do this or can we just leave it undefined?

hannessolo (approved):

nx2/blocks/chat/prompts/prompts.js
sharanyavinod: We could leave it undefined, but then would need to introduce fallbacks at multiple points within the code and this felt and cleaner and clearer to me.

PR #472, opened 2026-05-27
refactor(utils): centralize DA_AGENT URL via env resolver

anfibiacreativa:
Adds DA_AGENT to the shared getEnv-based service resolver so the agent URL follows the same ?da-agent=local / ?da-agent=ci / ?da-agent=prod convention as DA_ADMIN, DA_COLLAB, DA_CONTENT, etc.

Previously chat-controller.js hardcoded the agent URL with a ?ref=local ternary that pointed to port 4200 (out of date with the actual da-agent dev port 4002), causing repeated confusion when running the chat against the local agent. The centralized resolver eliminates the port mismatch and keeps service URLs in one place.

### DIFF
--- a/nx2/utils/utils.js
+++ b/nx2/utils/utils.js
@@ -40,6 +40,14 @@ const DA_LIVE_PREVIEW_ENVS = {
   prod: 'https://preview.da.live',
 };
 
+const DA_AGENT_ENVS = {
+  local: 'http://localhost:4002',
+  dev: 'http://localhost:4002',
+  stage: 'https://agent.da.live',
+  ci: 'https://da-agent-ci.adobeaem.workers.dev',
+  prod: 'https://agent.da.live',
+};
+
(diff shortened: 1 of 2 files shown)


nx2/utils/utils.js
```diff
@@ -40,6 +40,14 @@ const DA_LIVE_PREVIEW_ENVS = {
   prod: 'https://preview.da.live',
 };
 
+const DA_AGENT_ENVS = {
+  local: 'http://localhost:4002',
+  dev: 'http://localhost:4002',
+  stage: 'https://agent.da.live',
+  ci: 'https://da-agent-ci.adobeaem.workers.dev',
+  prod: 'https://agent.da.live',
```
mhaack, line 48: keep dev, stage & prod.

mhaack: @anfibiacreativa did you adjust this?

PR #475, opened 2026-05-28
# fix(chat): strip orphaned tool-call messages before sending to agent

anfibiacreativa:
# fix(chat): strip orphaned tool-call messages before sending to agent

## Summary
- Client-side defense against the Anthropic "tool_use ids without tool_result" rejection
- Replaces the old blanket filter (which nuked all non-approval tool-call messages) with a precise ID-matching approach that only drops genuinely orphaned ones

## Why
- The Anthropic API rejects any request where a `tool_use` block has no corresponding `tool_result` in the immediately following message
- Orphans appear when `streamText` hits its step-limit mid-tool-execution, or when the client strips virtual (non-approval) tool results from persisted history
- The previous filter was overly aggressive — it stripped all assistant array-content messages without an approval request, which hid legitimate completed tool calls from the conversation on reload

## What Changed
- **`nx2/blocks/chat/chat-controller.js`**
  - Added `stripOrphanedToolCallMessages()` — collects all resolved `toolCallId`s from `tool-result` messages, then drops any assistant message whose tool-call IDs are not in that set (unless it contains an approval request, which is resolved server-side)
  - Applied in `loadInitialMessages()` (cached history on page load) and `_stream()` (outbound payload to the agent) so orphans never reach the API
- **`.gitignore`** — added `PR-throwaways/`

## Test Plan
- [ ] Trigger a multi-tool skill (e.g. translate) and verify tool-call / approval / result flow works end-to-end
- [ ] Kill/restart the agent mid-tool-execution, reload, and confirm the chat recovers without the "tool_use ids" error
- [ ] Verify that completed approval sequences still render correctly on page reload (not stripped by the new filter)
- [ ] Run `npm test` in da-nx and confirm existing tests pass

## Risks / Follow-ups
- Depends on the `refactor/centralize-agent-url` PR (for the `DA_AGENT` import in chat-controller)
- The server-side counterpart (`ensureOrphanedToolResults` in da-agent) is a separate PR — together they form defense-in-depth, but each is independently sufficient
- If the approval-request check in `stripOrphanedToolCallMessages` ever gets out of sync with da-agent's `resolveApprovals`, approval messages could be incorrectly dropped — low risk since both use the same `TOOL_APPROVAL_REQUEST` constant

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ helix-importer-ui
 *.bak
 .idea
 .claude/settings.local.json
+PR-throwaways/
--- a/nx2/blocks/chat/chat-controller.js
+++ b/nx2/blocks/chat/chat-controller.js
@@ -6,6 +6,33 @@ import { DA_AGENT } from '../../utils/utils.js';
 
 const AGENT_URL = `${DA_AGENT}/chat`;
 
+/**
+ * Drop assistant array-content messages whose tool-call IDs have no matching
+ * tool-result anywhere in the history. These orphans appear when the agent's
+ * streamText step-limit fires mid-tool-execution or when the client strips
+ * virtual (non-approval) tool results. Without this filter the Anthropic API
+ * rejects the request with "tool_use ids without tool_result blocks".
+ */
+function stripOrphanedToolCallMessages(messages) {
+  const resolvedIds = new Set();
+  for (const msg of messages) {
+    if (msg.role === ROLE.TOOL && Array.isArray(msg.content)) {
+      for (const p of msg.content) {
+        if (p.type === AGENT_EVENT.TOOL_RESULT && p.toolCallId) resolvedIds.add(p.toolCallId);
+      }
+    }
+  }
+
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@ helix-importer-ui
 *.bak
 .idea
 .claude/settings.local.json
+PR-throwaways/
```
hannessolo, line 11: We can probably remove this?

hannessolo (approved):

.gitignore
anfibiacreativa: No, cause everytime I generate one I need to make sure it's not staged. I think it's fine for individuals to add their private items to .gitignore. I can move it under /private-docs if we establish a convention

hannessolo: It's just something that seems specific to your claude setup.

But I'm fine with establishing a convention - we can update agents.md to say "store throwaway PR documentation in /xyz" and then add xyz to gititnore.

PR #477, opened 2026-05-28
fix: add scope to agent change event

sharanyavinod:
- Adds scope of change ie file or document level + affected folder paths to the agent change event

Preview: https://ew-event--da-live--adobe.aem.live/canvas?nx=ew-event&nxver=2#/exp-workspace/frescopa/index 

Fixes: [SITES-45681: Tool ↔ UI refresh eventing](https://jira.corp.adobe.com/browse/SITES-45681)

### DIFF
--- a/nx2/blocks/chat/chat-controller.js
+++ b/nx2/blocks/chat/chat-controller.js
@@ -1,5 +1,5 @@
 import { loadIms } from '../../utils/ims.js';
-import { AGENT_EVENT, ROLE, TOOL_NAME, TOOL_STATE } from './constants.js';
+import { AGENT_EVENT, ROLE, TOOL_NAME, TOOL_SCOPE, TOOL_STATE } from './constants.js';
 import { readStream } from './utils.js';
 import { loadMessages, saveMessages, resetSession } from './persistence.js';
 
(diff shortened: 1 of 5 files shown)


nx2/blocks/chat/chat-controller.js
```diff
@@ -1,8 +1,23 @@
...
+  if (!org || !repo) return [];
+  const toParent = (p) => {
+    const parts = (p ?? '').replace(/^\//, '').split('/').filter(Boolean);
+    parts.pop();
+    return `/${org}/${repo}${parts.length ? `/${parts.join('/')}` : ''}`;
+  };
+  if (toolName === TOOL_NAME.CONTENT_MOVE) {
+    return [...new Set([toParent(input.sourcePath), toParent(input.destinationPath)])];
+  }
+  if (toolName === TOOL_NAME.CONTENT_COPY) return [toParent(input.destinationPath)];
```
hannessolo, line 17: I think hardcoding the tools for now is fine, at some point we might want to have the agent decide that it needs to be done. I guess we could have a tool like `REFRESH_FILES` that's a no-op on the server but also get's handled here.

But let's save that, and see if it's really needed.

hannessolo (dismissed):

hannessolo (approved):

PR #478, opened 2026-05-28
# feat(shared): promote nx-card and nx-tabs to nx2/blocks/shared

anfibiacreativa:
## Summary
- Adds `nx-card` and `nx-tabs` as first-class shared web components under
  `nx2/blocks/shared/`, alongside the existing `nx-popover`, `nx-dialog`,
  `nx-menu`, etc.
- No consumers in da-nx yet; this PR exists to make these primitives importable
  from the da.live CDN so external apps (ew-extensions Skills Editor, future
  da-live consumers) can stop carrying local copies.

## Why
- The ew-extensions Skills Editor maintains private copies of `nx-card`,
  `nx-tabs`, and `nx-popover`. The popover is literally a one-line-diff copy
  of `nx2/blocks/shared/popover/popover.js`.
- As part of the Skills Management redesign (see
  `private-docs/skills-storage-redesign-plan.md`), ew-extensions will switch
  to CDN-imported shared components via importmap, eliminating the duplicates.
- That switch is blocked until both primitives exist in da-nx and are live
  on the production CDN.

## What Changed
- **`nx2/blocks/shared/card/{card.js,card.css}`** — generic card primitive
  with `heading`, `subheading`, `pill`, `selected`, `interactive` props and
  `pill` / default / `actions` slots. Uses Spectrum 2 design tokens, supports
  light-dark color schemes.
- **`nx2/blocks/shared/tabs/{tabs.js,tabs.css}`** — accessible tab strip
  with `items` (`{id, label}[]`) and `active` props. Emits `tab-change` events.
  Full keyboard navigation (Arrow keys, Home, End) with focus management.
- Both follow the existing `nx2/blocks/shared/popover` conventions: `da-lit`
  imports, `loadStyle` via `import.meta.url`, plain `customElements.define`,
  shadow DOM with adopted stylesheets.

## Test Plan
- [ ] `npm run lint:js` and `npm run nx2:lint:css` are clean (verified)
- [ ] After deploy, `https://da.live/nx2/blocks/shared/card/card.js` and
      `https://da.live/nx2/blocks/shared/tabs/tabs.js` resolve as ES modules
- [ ] Smoke-test in a scratch HTML page (importmap → `da-lit` → render
      `<nx-card heading="..." />` and `<nx-tabs .items=${[...]} />`)
- [ ] Visual parity with the current ew-extensions copies (Skills Editor
      side-by-side comparison once PR-6 lands)

## Risks / Follow-ups
- No active consumers yet — risk is limited to "deploy adds dead code".
- Follow-up: ew-extensions PR-6 (Skills Editor importmap switch) consumes
  these via `https://da.live/nx2/blocks/shared/...`.
- Future: `da-live` browse/canvas/nav surfaces are candidates for adoption
  once the patterns prove out in the Skills Editor.

### DIFF
--- a/nx2/blocks/shared/card/card.css
+++ b/nx2/blocks/shared/card/card.css
@@ -1,0 +1,86 @@
+:host {
+  display: block;
+  font-family: var(--s2-font-family);
+}
+
+.card {
+  display: flex;
+  align-items: center;
+  gap: var(--s2-spacing-100);
+  padding: var(--s2-spacing-75) var(--s2-spacing-100);
+  border: 1px solid var(--s2-gray-200);
+  border-radius: var(--s2-corner-radius-200);
+  background: light-dark(var(--s2-gray-25), var(--s2-gray-50));
+  transition: border-color 0.15s, background 0.15s;
+  user-select: none;
+}
+
+:host([selected]) .card {
+  border-color: var(--s2-blue-900);
+  background: light-dark(var(--s2-blue-100), var(--s2-blue-200));
+}
+
+:host([interactive]) .card {
+  cursor: pointer;
+}
+
+:host([interactive]) .card:hover {
+  border-color: var(--s2-gray-400);
+  background: light-dark(var(--s2-gray-50), var(--s2-gray-75));
+}
+
+/* ─── left pill/badge slot ───────────────────────────────────────────────── */
+
+.card-pill {
+  flex-shrink: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 1.75rem;
+  height: 1.75rem;
+  border-radius: var(--s2-corner-radius-200);
+  background: var(--s2-gray-200);
+  font-size: 10px;
+  font-weight: var(--s2-heading-font-weight);
+  text-transform: uppercase;
+  letter-spacing: 0.03em;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+/* ─── body ───────────────────────────────────────────────────────────────── */
+
+.card-body {
+  flex: 1;
+  display: flex;
+  flex-direction: column;
+  gap: 1px;
+  min-width: 0;
+}
+
+.card-heading {
+  font-size: var(--s2-body-size-s);
+  font-weight: var(--s2-heading-font-weight);
+  color: var(--s2-gray-900);
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.card-subheading {
+  font-size: var(--s2-body-size-xs);
+  color: var(--s2-gray-600);
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+/* ─── actions / checkbox slot ────────────────────────────────────────────── */
+
+.card-actions {
+  display: flex;
+  align-items: center;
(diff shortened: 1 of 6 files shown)


mhaack:
More a general question: do we have UX for the two components? Are they cross used? Not sure if we should have these in da-nx yet.

anfibiacreativa:
> * No active consumers yet — risk is limited to "deploy adds dead code".

This is a very good question and I describe it in risks in the PR risks
I think it is a good way to move forward. As for having UX, nope. We need @kamendola in the loop

sharanyavinod:
@anfibiacreativa Not sure I understand - why can't extensions reuse popover from da-nx without moving card and tabs into shared? Neither of those have usage outside of of skills yet.

anfibiacreativa:
> @anfibiacreativa Not sure I understand - why can't extensions reuse popover from da-nx without moving card and tabs into shared? Neither of those have usage outside of of skills yet.

We absolutely can, the rationale here is: we have identified these components as shared components because we know sooner or later we will need them somewhere else (cards and tabs are typically a core/common to all design systems).

If those design primitives stay in ew they're not discoverable and the moment a consumer appears, they may never know they exist in ew-extensions. Instead of backporting them at that time, they will likely recreate them. 

Architecturally we face a trade off here, and I'm bullish that this is the right path. zero active consumers = zero migration cost

CC: @mhaack

sharanyavinod:
Discoverability isn't the right argument here though, since we own the design and would define when other surfaces adopt these components. We would then know to make the switch if needed. Without mocks we can't know if the contract is right, and locking it in before a second consumer exists is a breaking change risk we'd be taking on speculatively. Precedent: pill lives in chat for exactly this reason. IMO we should keep card and tabs in ew-extensions until we have designs and a real second use case.

mhaack:
If these two components are at the moment only used for the skills management tool, why not keep them there? We can promote later when we have a reviewed UX and shared usage.

anfibiacreativa:
> Discoverability isn't the right argument here though, since we own the design and would define when other surfaces adopt these components.

I don't really think there is right or wrong here. Yours is a fair point, but I think the framing is closer to a proprietary codebase with a stable team, while this is an OSS library deployed to a public CDN. The bar for what belongs in a shared lib is important.

"We'll know to promote it when we need it" works when you own the roadmap and the team. But hopefully here, contributors come and go, and in OSS the discovery problem is real. The moment a second consumer appears and doesn't know nx-card exists in ew-extensions, you get a duplicate. we've all seen this play out.

On the API risk: I get it for something domain-specific, but card and tabs are fairly simple. heading, a few slots, an items array. The API surface is near-zero. The pill precedent makes sense precisely because pill has domain behavior that's hard to generalize without a design. these don't.

So yes, looping in @kamendola is the right call, and I'm committed to not adding consumers until we have UX sign-off. But I don't think we should wait to merge this. The risk of being wrong in either direction is negligible for these two components, and leaving them in ew-extensions creates more risk over time, not less.

hannessolo:
I agree with not moving it into nx yet. Nx is being used in production, and since these components don't have the "correct" styling yet it seems weird to me to merge them.

anfibiacreativa:
> I agree with not moving it into nx yet. Nx is being used in production, and since these components don't have the "correct" styling yet it seems weird to me to merge them.

Ok, let's prioritize the design of whatever we consider core components, then. So we can make sure to abstract anything that's a common

PR #483, opened 2026-05-29
fix: markdown rendering for custom types

sharanyavinod:
- Handles rendering of all custom types defined here - https://github.com/adobe-rnd/da-agent/blob/main/docs/agent-markdown-outputs.md - except for ones related to skills
- Note: no final mocks yet, so basic styling done using spectrum for inspiration

Preview at https://da.live/canvas?nx=ew-md&nxver=2#/exp-workspace/frescopa/index


nx2/blocks/chat/renderers.js
```diff
@@ -48,10 +48,98 @@ function renderNode(node) {
...
+  if (type === 'toggle-list') return renderToggleList(tree);
+  if (type === 'checklist') return renderChecklist(tree);
+  return html`<div class="directive directive-${type}">${renderNode(tree)}</div>`;
+}
+
 function renderMessageContent(text) {
   if (!text) return nothing;
-  const tree = parser.parse(text);
-  return renderNode(tree);
+  const segments = parseDirectives(text);
```
anfibiacreativa, line 105: parseDirectives is a pure text-parsing function (no DOM, no Lit, no side effects) living in a renderer module. per our convention, pure utilities belong in nx2/utils/. it's already exported and has tests, so moving it should be low risk low-risk.

nx2/blocks/chat/chat.css
```diff
@@ -192,6 +193,104 @@
...
+
+    &:checked {
+      background: var(--s2-gray-800);
+
+      &::after {
+        content: "";
+        position: absolute;
+        inset: 0;
+        background-color: var(--s2-gray-25);
+        mask-image: url("/nx2/img/icons/S2_Icon_Checkmark_20_N.svg");
```
anfibiacreativa: mask-image: url("/nx2/img/icons/S2_Icon_Checkmark_20_N.svg") is a hardcoded absolute asset path. breaks if the asset structure ever moves. we have flagged this before  (see other icon usage with loadHrefSvg). @hannessolo  what was the decision here?

nx2/blocks/chat/chat.css
```diff
@@ -192,6 +193,104 @@
...
+    align-items: flex-start;
+    gap: var(--s2-spacing-75);
+  }
+
+  input[type="checkbox"] {
+    position: relative;
+    width: 14px;
+    height: 14px;
+    flex-shrink: 0;
+    margin-top: 4px;
```
anfibiacreativa: we're using var(--s2-spacing-75) right next to these. do spacing tokens cover 4px and 6px? if so, let's use them too.

nx2/blocks/chat/renderers.js
```diff
@@ -48,10 +48,98 @@ function renderNode(node) {
...
+  const segments = [];
+  const buf = [];
+  let type = null;
+  let openLine = null;
+
+  for (const line of text.split('\n')) {
+    if (!type && line.startsWith(':::')) {
+      const content = buf.splice(0).join('\n');
+      if (content) segments.push({ kind: 'text', content });
+      [type] = line.slice(3).split(' ');
```
anfibiacreativa: what happens with a bare ::: with no type name? type ends up as an empty string and we get class="directive directive-". worth a guard or at least a test case.

nx2/blocks/chat/chat.css
hannessolo: Yes, we should use `<svg><use>` wherever possible. The icons shouldn't be added to the code, but referenced directly from da.live, eg. https://da.live/img/icons/s2-icon-checkmark-20-n.svg

sharanyavinod: This is pure styling, so didnt want to add an svg. I'll update to have the live url then - I saw a couple other similar usages, will adapt those in parallel as well.

hannessolo: Yes, that's why "wherever possible" - I think here your approach might make more sense. But I'm not sure, will dark mode work with mask-image?
In either case we should use the URL from da.live instead of adding the svg to the code.

sharanyavinod: Yes, currentcolor on the background will inherit the parent color. I'll adapt to use the url -> for the others I think we might need to first add the icons to da live

hannessolo:
Nice, looks much better!

I'm fine with the current implementation for now, but as an improvement we should find a way to already render the correct style before the closing tag appears.

sharanyavinod:
> we should find a way to already render the correct style before the closing tag appears

Done :)

hannessolo (approved):

nx2/blocks/chat/chat.css
```diff
@@ -192,6 +193,105 @@
...
+    &:checked {
+      background: var(--s2-gray-800);
+      color: var(--s2-gray-25);
+
+      &::after {
+        content: "";
+        position: absolute;
+        inset: 0;
+        background-color: currentcolor;
+        mask-image: url("https://da.live/img/icons/s2-icon-checkmark-20-n.svg");
```
anfibiacreativa, line 266: can we set the URL via a JS custom property (--checkmark-icon: url(${...})) using codeBase, the same way inline SVG <use> refs are handled instead of hardcoding the CDN url?

nx2/blocks/chat/utils/icons.js
```diff
@@ -0,0 +1,10 @@
+const IMAGE_EXTS = new Set(['png', 'jpg', 'jpeg', 'gif', 'webp', 'svg', 'mp4', 'webm', 'mov']);
+const TABLE_EXTS = new Set(['json', 'xlsx', 'xls', 'csv']);
+
+export function fileIconName(filename) {
```
anfibiacreativa, line 4: the old code had a comment saying this mirrors entryTypeFromExtension in browse/utils.js. we removed the comment but the duplication is still there. why not move fileIconName to nx2/utils/ and delete browse/utils.js's version, or at least import from there? two parallel implementations will diverge.

sharanyavinod: Browse doesnt exist anymore - the comment is a remnant from when we had custom browse block similar to chat.

nx2/blocks/chat/chat.css
sharanyavinod: We could, but mask-image on a pseudo-element is purely decorative styling with no functional JS dependency. Adding a custom property would couple js and css just to avoid a stable CDN URL and the trade-off didnt feel worth it here.

anfibiacreativa (approved):
LGTM!

PR #489, opened 2026-06-02
fix(chat): wire up Skills Editor prompt + and ▶ buttons

anfibiacreativa:
The Skills Editor (ew-extensions) prompt cards dispatch `prompt-add` / `prompt-send` on a `da-skills-editor` BroadcastChannel and as window CustomEvents, but `nx-chat` never subscribed — so clicking + or ▶ was a no-op on the chat side. Add a small bridge in connectedCallback that listens on both transports, dedupes the editor's double-dispatch, and routes prompt-add to fill the input only and prompt-send to fill + submit.

### DIFF
--- a/nx2/blocks/chat/chat.js
+++ b/nx2/blocks/chat/chat.js
@@ -26,6 +26,22 @@ const icon = (name) => html`<svg class="chat-icon" viewBox="0 0 20 20" aria-hidd
 
 const UI_PROMPTS_GAP = 8;
 
+// Cross-window bridge from the Skills Editor (ew-extensions) prompt cards.
+// The editor's `+` (add) and `▶` (run) buttons post `prompt-add` / `prompt-send`
+// on this BroadcastChannel and as window CustomEvents.
+const SKILLS_CHANNEL = 'da-skills-editor';
+const SKILLS_PROMPT_ADD_EVENTS = [
+  'da-skills-editor-prompt-add-to-chat',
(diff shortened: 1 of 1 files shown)


nx2/blocks/chat/chat.js
```diff
@@ -27,6 +26,22 @@ const icon = (name) => html`<svg class="chat-icon" viewBox="0 0 20 20" aria-hidd
 
 const UI_PROMPTS_GAP = 8;
 
+// Cross-window bridge from the Skills Editor (ew-extensions) prompt cards.
+// The editor's `+` (add) and `▶` (run) buttons post `prompt-add` / `prompt-send`
+// on this BroadcastChannel and as window CustomEvents.
+const SKILLS_CHANNEL = 'da-skills-editor';
+const SKILLS_PROMPT_ADD_EVENTS = [
```
hannessolo, line 33: I don't think this code should know about the skills editor. We should just have a generic event that we can send on the chat component that sets the prompt.

sharanyavinod: +1 - I would look into extending the existing add-to-chat event to also accept prompt instead

hannessolo: I have a PR for that here: https://github.com/adobe/da-nx/pull/490

mhaack: +1 we should proceed with the generic approach which can be used by others as well

PR #495, opened 2026-06-04
fix: simplify toggle list rendering

sharanyavinod:
- Remove leftover logic for rendering toggle list within details+summary and use simple ul+li. Resolves flickering for custom toggle-list response from chat.

Preview: https://da.live/canvas?nx=ew-list&nxver=2#/exp-workspace/frescopa/index


nx2/blocks/chat/renderers.js
```diff
@@ -63,23 +63,9 @@ function renderChecklistItem(node) {
...
-        body.push(renderNode(tree.children[i]));
-      }
-      items.push(html`<li>${summary}<p>${body}</p></li>`);
-    } else {
-      items.push(html`<li>${renderNode(node)}</li>`);
-    }
-    i += 1;
-  }
+  const items = tree.children.map((node) => (node.type === 'blockquote'
+    ? html`<li>${node.children.map(renderNode)}</li>`
```
anfibiacreativa, line 67: we're dropping the  `<li> `wrapping for non-blockquote nodes here. they'll be bare children of the `<ul>,` which is invalid HTML. was this intentional?

anfibiacreativa: also, the old code collected non-blockquote nodes after each blockquote as the toggle body content. this simplification drops that grouping entirely. if toggle items are supposed to have summary+body, this is a regression. can we verify with a toggle list that has body content?

nx2/blocks/chat/renderers.js
```diff
@@ -63,23 +63,9 @@ function renderChecklistItem(node) {
 }
 
 function renderToggleList(tree) {
-  const items = [];
-  let i = 0;
-  while (i < tree.children.length) {
-    const node = tree.children[i];
-    if (node.type === 'blockquote') {
-      const summary = node.children.flatMap((c) => (c.children ?? []).map(renderNode));
```
anfibiacreativa, line 71: the old code did flatMap to strip the paragraph layer from blockquote children (so you'd get inline content in the <li>). map(renderNode) keeps the <p> wrapping. the rendered HTML structure changes here. is that tested?

anfibiacreativa (changes requested):

anfibiacreativa (approved):
Approved after problem clarification. Pending opening ticket for agent

PR #501, opened 2026-06-08
fix: message formatting

sharanyavinod:
- Fix message formatting
  - Blockquote was stripped
  - Tables not formatted -> reusing remarkGfmNoLink as in [da-nx](https://github.com/adobe/da-nx/blob/main/nx/deps/mdast/src/index.js#L15)
  - Using mdast -> hast transformation for rendering to avoid custom logic for determining which html elements to use for which mdast types.  No new imports introduced for these, already used within da-nx for the editor. 
  - Introduces dep for converting hast -> dom: avoids having to explicitly specify rendering logic for each element. However, this is a new dep vs existing hast -> html in the repo over which this was preferred to avoid injecting html string as innerhtml via lit's unsafehtml directive
- Fix message flickering on scroll

Preview: https://da.live/canvas?nx=ew-chat&nxver=2#/exp-workspace/frescopa/index


nx2/blocks/chat/renderers.js
```diff
@@ -6,94 +6,88 @@ import { fileIconName } from './utils/icons.js';
...
+    case 'blockquote': return html`<blockquote>${children()}</blockquote>`;
+    case 'table': return html`<table>${children()}</table>`;
+    case 'thead': return html`<thead>${children()}</thead>`;
+    case 'tbody': return html`<tbody>${children()}</tbody>`;
+    case 'tr': return html`<tr>${children()}</tr>`;
+    case 'th': return html`<th>${children()}</th>`;
+    case 'td': return html`<td>${children()}</td>`;
+    case 'a': {
+      const href = SAFE_URL.test(node.properties?.href) ? node.properties.href : '#';
+      return html`<a href="${href}" target="_blank" rel="noopener noreferrer">${children()}</a>`;
```
mhaack: We don't have all hast utils, but https://github.com/syntax-tree/hast-util-from-html should give us hast > html rendering. We maybe consider this.

sharanyavinod: Good call.I did debate about this a bit - the concern was that if we use such a util to convert hast to html string, we would then have to render via lit's unsafehtml which uses innerhtml. 
But looks like theres a different util which converts hast to dom instead of html which would avoid the html string injection concerns. Let me try that one instead.

sharanyavinod: @mhaack Works as expected and is clean with https://github.com/syntax-tree/hast-util-to-dom -> however this adds a new dependency whereas to-html exists but that has the innerhtml concern as mentioned. Repo states to use existing deps wherever possible, but in this case where input is not controlled by us, I tend to lean towards adding the new package instead. Wdyt?

mhaack: In general, I like the approach because we're not reinventing something which already exists. However, I wonder if we can still simplify the way we apply `toDom` for nested lists and other structures.

mhaack (approved):

PR #503, opened 2026-06-08
feat: add cashebust bool param

chrischrischris:


auniverseaway (approved):

auniverseaway:
A `no-store` header is likely cleaner, but might take a bit more work to validate.

PR #505, opened 2026-06-10
fix :  401 Unauthorized when fetching pages behind site authentication for import

kmurugulla:
Summary of Changes

  **Problem**

  Importer tool failed with 401 Unauthorized when fetching pages behind AEM site authentication (requiring Sidekick login).

  **Root Cause**

  The Importer was using IMS bearer tokens directly instead of exchanging them for site-specific authentication tokens required by protected AEM sites.

  **Solution**

  Implemented AEM site token exchange mechanism :

  ---
  **How It Works**

  1. Token Exchange: First request exchanges IMS token for site-specific token (one per org/site)
  2. Caching: Site token cached per org/site/ref to avoid repeated exchanges
  3. Authentication: Uses Authorization: token ${siteToken} header format
  4. Retry Logic: On 401/403, clears cache and retries with fresh token once
  5. Expiry Handling: Tokens auto-cleared when within 60s of expiration

  ---
  **Testing**

  Tested with protected site main--vg-macktrucks-com--volvogroup.aem.page:
  - ✅ Token exchange successful
  - ✅ Pages behind authentication now import successfully
  - ✅ Token properly cached across multiple page imports
  - ✅ Retry logic works on auth failures

### DIFF
--- a/nx/blocks/importer/importer.js
+++ b/nx/blocks/importer/importer.js
@@ -85,8 +85,11 @@ class NxImporter extends LitElement {
     this._urls = [];
 
     if (data.index) {
-      const opts = getOptions();
-      const { origin } = new URL(data.index);
+      const indexUrl = new URL(data.index);
+      const { origin } = indexUrl;
+      // Parse source org/repo from the query index URL hostname
+      const [fromRepo, fromOrg] = indexUrl.hostname.split('.')[0].split('--').slice(1).slice(-2);
+      const opts = await getOptions(fromOrg, fromRepo);
       const proxyUrl = `https://da-etc.adobeaem.workers.dev/cors?url=${encodeURIComponent(data.index)}`;
       const resp = await fetch(proxyUrl, opts);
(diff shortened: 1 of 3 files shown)


auniverseaway (dismissed):
There's an opportunity to move the getSiteToken code into `nx2/utils/api.js`. This is definitely **_ok_** as-is, but **_great_** would be a those bits moved.

https://github.com/adobe/da-live/blob/main/blocks/shared/utils.js#L282-L303 <-- you can copy this function and it will do everything for you. I think there's some simplification you can get out of it, too.

kmurugulla:
> There's an opportunity to move the getSiteToken code into `nx2/utils/api.js`. This is definitely **_ok_** as-is, but **_great_** would be a those bits moved.
> 
> https://github.com/adobe/da-live/blob/main/blocks/shared/utils.js#L282-L303 <-- you can copy this function and it will do everything for you. I think there's some simplification you can get out of it, too.

Thank you @auniverseaway . Made the changes and tested. please review

auniverseaway (approved):

PR #506, opened 2026-06-11
ImageLoc: Throttle GLaaS multimodal API calls and avoid status probe storms after failed cancel.

janaki-r-bhagwath:
- Share one rolling 100/min limiter across getPutURLForFile and getV2Asset (headroom below documented 120/min)
- Retry getPutURL 429s using Retry-After / X-Rate-Limit-Retry-After-Seconds (+250ms), with window-based fallback
- Split image send into parallel DA fetch (5) then throttled GLaaS upload (3)
- Queue v2 status probes (concurrency 3) and poll locales sequentially instead of firing all assets × langs at once
- Return cancel success from updateStatus; skip handleGetStatus when GLaaS rejects cancel (e.g. Invalid Status 400)
- Harden getPutURL retry tests and reset shared limiter in v2 status tests

### DIFF
--- a/nx/blocks/loc/connectors/glaas/multimodalApi.js
+++ b/nx/blocks/loc/connectors/glaas/multimodalApi.js
@@ -184,17 +184,6 @@ export async function fetchBlobFromSignedUrl(signedURL) {
 
 const CONTENT_DA_LIVE = 'content.da.live';
 
-/** One srcset candidate URL; strips trailing width/density descriptor (e.g. 600w, 2x) only. */
-export function parseSrcsetUrl(part) {
-  const trimmed = part.trim();
-  if (!trimmed) return '';
-  return trimmed.replace(/\s+\d+(?:\.\d+)?[wx]\s*$/i, '').trim();
-}
-
-function collectSrcsetUrls(srcset) {
-  return srcset.split(',').map(parseSrcsetUrl).filter(Boolean);
-}
-
 /** Encode delivery URL for HTML src/srcset (spaces → %20, valid srcset). */
 export function contentDaLiveHrefForAttribute(href) {
   if (!href) return href;
@@ -225,18 +214,13 @@ function isProjectContentDaLiveUrl(href, org, site) {
   }
 }
 
-/** MVP: absolute https://content.da.live/... image URLs only (not relative ./media_ from DNT). */
+/** MVP: absolute https://content.da.live/... image URLs from img[src] only (not relative ./media_ from DNT). */
 export function collectContentDaLiveImageUrls(html, { org, site } = {}) {
   const doc = new DOMParser().parseFromString(html, 'text/html');
   const urls = new Set();
   doc.querySelectorAll('img[src]').forEach((img) => {
     const src = img.getAttribute('src');
     if (isProjectContentDaLiveUrl(src, org, site)) urls.add(new URL(src).href);
-  });
-  doc.querySelectorAll('source[srcset]').forEach((source) => {
-    collectSrcsetUrls(source.getAttribute('srcset') || '').forEach((src) => {
-      if (isProjectContentDaLiveUrl(src, org, site)) urls.add(new URL(src).href);
-    });
   });
   return [...urls];
 }
@@ -258,19 +242,6 @@ export function contentDaLivePathKey(href) {
   }
 }
 
-function replaceSrcsetUrls(srcset, resolveNewUrl) {
-  return srcset.split(',').map((part) => {
-    const trimmed = part.trim();
-    if (!trimmed) return part;
-    const src = parseSrcsetUrl(trimmed);
-    const descriptor = trimmed.slice(src.length).trim();
-    const resolved = resolveNewUrl(src);
-    if (!resolved) return part;
-    const encoded = contentDaLiveHrefForAttribute(resolved);
-    return descriptor ? `${encoded} ${descriptor}` : encoded;
-  }).join(', ');
-}
-
 /** Replace content.da.live image URLs using pathname → new delivery URL map. */
 export function rewriteContentDaLiveImageUrls(html, pathToNewUrl) {
   const doc = new DOMParser().parseFromString(html, 'text/html');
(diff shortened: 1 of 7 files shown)


nx/blocks/loc/connectors/glaas/multimodalApi.js
```diff
@@ -327,16 +488,53 @@ export function v2AssetStatusFromProbe(assetName, meta) {
...
+  return { results };
+}
+
 async function probeMultimodalAssetStatuses({
   service, token, task, langCode, assetNames,
 }) {
   const langTask = { ...task, code: langCode };
-  const probes = await Promise.all(
-    assetNames.map(async (assetName) => {
+  const { results: probes } = await runImageQueue({
```
jenssingler: In case of an error in `runImageQueue` the destructuring of the result leaves `probes` as `undefined`. That means on the caller site (line 556) will be `undefined` and will throw in line 563.

nx/blocks/loc/views/translate/translate.js
```diff
@@ -218,22 +218,28 @@ class NxLocTranslate extends LitElement {
 
     const { cancelTranslation } = this._service.connector;
 
+    let cancelAccepted = true;
     for (const lang of this._translateLangs) {
-      await cancelTranslation({ service: this._service, lang, sendMessage });
+      const result = await cancelTranslation({ service: this._service, lang, sendMessage });
+      if (result?.ok === false) cancelAccepted = false;
```
jenssingler: I think this will fail for multiple lang, if  one of them is failing. Meaning `handleGetStatus` is never called and the languages will show "canceling" even if some of them did complete successfully.

nx/blocks/loc/connectors/glaas/multimodalApi.js
```diff
@@ -4,7 +4,154 @@ import { daFetch } from '../../../../utils/daFetch.js';
...
+const IMAGE_UPLOAD_QUEUE_CONCURRENCY = 3;
+const V2_PROBE_QUEUE_CONCURRENCY = 3;
+const IMAGE_PUSH_INTERVAL_MS = 250;
+const PUT_URL_MAX_RETRIES = 4;
+const PUT_URL_RETRY_WAIT_MS = 1000;
+const PUT_URL_429_FALLBACK_DELAY_MS = Math.ceil(GLAAS_API_WINDOW_MS / 2) + 250;
+
+let glaasApiRateLimitUntil = 0;
+
+function delay(ms) {
```
jenssingler: You could just export `throttle` from api.js and reuse it here.

jenssingler (changes requested):
I cannot play it through right now, but reading the code and I see there is a potential of wasted time and penalizing unrelated calls with `30250ms` delay (via `glaasApiRateLimitUntil`), if the rate-limit headers are missing and the rolling limit is already nearly full (`windowRetryDelayMs`)
E.g. `getPutUrlForFile` gets 429 and no header -> 30250ms delay. `acquireGlaasApiSlot` then blocks other calls, e.g. `getV2Asset`.

I would think this gets more pronounced if multiple tabs and users are performing a similar operation.

nx/blocks/loc/connectors/glaas/multimodalApi.js
```diff
@@ -148,9 +327,15 @@ export async function getV2Asset(service, token, task, assetName) {
   const [product = '', project = ''] = workflow?.split('/') ?? [];
   const opts = getOpts(clientid, token);
   try {
+    await acquireGlaasApiSlot();
```
jenssingler, line 330: I think the way this is written, the parallelism is gone. Since `acquire` is waiting 600ms, a simple status call (e.g. `probeMultimodelAssetStatuses`) is waiting for its slot. With 5 assets and 30 langs, this is 150x600ms = 90 seconds. 

Is there some indication to the user, that the call is taking time to avoid triggering this multiple times?

nx/blocks/loc/views/translate/translate.js
```diff
@@ -218,22 +218,28 @@ class NxLocTranslate extends LitElement {
...
+    let shouldRefresh = false;
     for (const lang of this._translateLangs) {
-      await cancelTranslation({ service: this._service, lang, sendMessage });
+      const result = await cancelTranslation({ service: this._service, lang, sendMessage });
+      if (result?.ok !== false) shouldRefresh = true;
     }
 
-    // Re-fetch status to ensure the service canceled everything.
-    this.handleGetStatus();
+    if (shouldRefresh) {
```
jenssingler, line 227: A lang can have multiple tasks. So I'm not sure if you should fetch status only if all cancel succeed. 
Is handleGetStatus "expensive" ?

nx/blocks/loc/connectors/glaas/multimodalApi.js
```diff
@@ -45,21 +184,61 @@ export function logMultimodalRequest(step, detail) {
...
-    return { error: 'Error getting put URL for file.' };
+
+  let waitInterval = PUT_URL_RETRY_WAIT_MS;
+  for (let attempt = 0; attempt <= maxRetries; attempt += 1) {
+    try {
+      await acquireGlaasApiSlot();
+      const resp = await fetch(url, opts);
+      if (resp.status === 429 && attempt < maxRetries) {
+        waitInterval *= 2;
+        const waitMs = putUrl429RetryDelayMs({ resp, waitInterval });
```
jenssingler, line 207: If the serves responds with "retry in 1 sec", then the waitMs is 1250ms... so there is no more exponential back off.
I guess for some cases it's ok, but I don't think that was the intention.

nx/blocks/loc/connectors/glaas/multimodalApi.js
```diff
@@ -45,21 +184,61 @@ export function logMultimodalRequest(step, detail) {
...
-    if (!json.putURL) return { error: 'Missing putURL in response.', status: resp.status, json };
-    logRequest?.('getPutURL-response', { status: resp.status, assetName });
-    return { putURL: json.putURL, instanceId: json.instanceId, status: resp.status };
-  } catch {
-    return { error: 'Error getting put URL for file.' };
+
+  let waitInterval = PUT_URL_RETRY_WAIT_MS;
+  for (let attempt = 0; attempt <= maxRetries; attempt += 1) {
+    try {
+      await acquireGlaasApiSlot();
```
jenssingler, line 203: For the 429 case you know how long you have to wait but on the second loop, you still acquire a slot. Not sure, if that's intended.

jenssingler (approved):
Found a couple of things that seemed odd to me... but approving since I'm not sure if they are really an issue.

PR #507, opened 2026-06-12
refactor: improve chat

sharanyavinod:
Note: Bulk of the changes are within the controllers and messages.js -> rest are either minor improvements or extracting logic from the controller for better isolation

**Approval flow rewrite**
The previous controller used a simple boolean _thinking flag and handled approvals serially. This PR replaces it with a proper turn lifecycle (idle → streaming → approval-pending → resuming) and a robust approval model:
- Per-tool streaming: each approval fires its own stream call immediately. Batching all approvals and sending them together causes the agent to execute tools internally without streaming results back - the client never receives them and can't build proper tool_result messages in history.
- Concurrent approval guard: _activeApprovalStreams counter (not boolean) prevents a completed stream from ending the turn while a second approval stream started by a rapid click is still open.
- Eager message commit: approval-request messages are committed to history the moment the event arrives, not deferred to stream end. Without this, clicking "Approve" before the stream closes orphans the response and causes a prefill error on the next call.
- buildAgentMessages: strips previous turns' raw tool call/result pairs from the agent payload (agent doesn't need them and keeping them caused "tool_use without tool_result" errors). Current turn context is always preserved.
- stripOrphanedToolCallMessages hardened: stale approval-responses and their matching approval-requests are both stripped when the tool result exists, and abandoned sequences (approval complete but no result and not live) are stripped to prevent orphaned tool_use blocks reaching agent.
- Path-based duplicate detection: when the model generates two calls for the same tool+path in one step, the second is blocked along with its approval request so it doesn't corrupt history.
- Persist on failure: _persistMessages() called at the end of every stream (success or error) so the user's input and any error message survive a page reload.
- finishReason: 'tool-calls' continuation: auto-continues the conversation when da-agent signals more tool calls without requiring another user message.

Other minor improvements:
- Extract utils to standalone files + add unit tests
- resolve chat messages flickering
- tool cards fallback to show input if available on expand, if no info available simply showing the tool name and disabling expand
- add missing icons to chat menu
- styling updates as per updated designs

Preview: https://da.live/canvas?nx=ewchat#/exp-workspace/frescopa/drafts/svinod/test


nx2/blocks/chat/controllers/chat-controller.js
```diff
@@ -0,0 +1,539 @@
...
+        ...(typeof a.sizeBytes === 'number' ? { sizeBytes: a.sizeBytes } : {}),
+      };
+    });
+  }
+
+  // Persists current messages. Called at stream end (success or error) so that the user's
+  // input and any error message survive a page reload even when no text was streamed.
+  _persistMessages() {
+    if (!this._messages) return;
+    this._getRoom().then((room) => saveMessages(room, this._messages, this._sessionId));
```
anfibiacreativa: _persistMessages() calls _getRoom().then(...) with no .catch(). if _getRoom() or saveMessages throws (IMS down, storage full), you get an unhandled rejection silently.

can we add a .catch(() => {}) at minimum here? silent data loss on page reload when persistence fails is worse than a logged warning.

nx2/blocks/chat/controllers/chat-controller.js
```diff
@@ -0,0 +1,539 @@
...
+    this._turn.resume();
+    this._update();
+    this._stream().catch((err) => {
+      if (err.name !== 'AbortError') {
+        this._messages = [
+          ...this._messages,
+          { role: ROLE.ASSISTANT, content: `Error: ${err.message}` },
+        ];
+      }
+    }).finally(() => this._done());
```
anfibiacreativa: looks like a missing this._persistMessages() in _continueTurn's .finally(). as I see, sendMessage and approveToolCall both do it.

nx2/blocks/chat/controllers/chat-controller.js
```diff
@@ -0,0 +1,539 @@
...
+    if (!this._messages) return;
+    this._getRoom().then((room) => saveMessages(room, this._messages, this._sessionId));
+  }
+
+  _update() {
+    this._onUpdate({
+      messages: this._messages,
+      thinking: this._turn.isActive,
+      streamingText: this._streamingText,
+      connected: this._connected,
```
anfibiacreativa: _connected → _isConnected, _pendingContinuation → _hasPendingContinuation. 

etc. we are trying to enforce this convention for booleans

nx2/blocks/chat/utils/tools.js
```diff
@@ -0,0 +1,17 @@
+import { TOOL_NAME } from '../constants.js';
+
+// Derives the parent folders affected by a tool call so the file browser can refresh.
+export function affectedFolders(toolName, input) {
+  const { org, repo } = input ?? {};
+  if (!org || !repo) return [];
+  const toParent = (p) => {
```
anfibiacreativa: can we rename to filePath or just path. toParent(path) reads cleaner.

nx2/blocks/chat/renderers.js
```diff
@@ -34,33 +34,35 @@ function renderMessageContent(text) {
...
+  const status = failed ? html`<span class="tool-card-status">${state}</span>` : nothing;
+  return detail ? html`
     <details class="tool-card tool-card-${state}">
-      <summary>${toolName}${failed ? html`<span class="tool-card-status">${state}</span>` : nothing}</summary>
-      ${detail ? html`<span class="tool-card-detail">${detail}</span>` : nothing}
-    </details>
-  `;
+      <summary>${toolName}${status}</summary>
+      <span class="tool-card-detail">${detail}</span>
+    </details>` : html`<span class="tool-card tool-card-detail">${toolName}${status}</span>`;
```
anfibiacreativa: the span in the no-detail fallback has both tool-card and tool-card-detail on it. is the tool-card class intentional there? if we need both styles to apply, we'd normally nest the elements, not stack the classes.

nx2/blocks/chat/constants.js
```diff
@@ -8,18 +8,14 @@ const MENU_OPTIONS = {
 
 const ADD_MENU_ITEMS = [
   { section: 'Add' },
-  { id: 'files', label: 'Files or images', icon: 'link' },
-  { id: MENU_OPTIONS.PROMPT, label: 'Prompt' },
-  { id: 'command', label: '"/" Command' },
+  { id: MENU_OPTIONS.FILES, label: 'Files or images', icon: 'link' },
+  { id: MENU_OPTIONS.PROMPT, label: 'Prompt', icon: 'commentremove' },
```
anfibiacreativa, line 12: can you confirm MENU_OPTIONS.FILES and MENU_OPTIONS.COMMAND are defined in the full MENU_OPTIONS object? the diff doesn't show them and the old code used string literals 'files' and 'command'.

nx2/blocks/chat/constants.js
sharanyavinod, line 7: @anfibiacreativa Yes, the values were here, just never used within menu items

nx2/blocks/chat/controllers/chat-controller.js
sharanyavinod: I don't think we have a consistent pattern yet, but I can adapt to use isConnected.
However, with the other one, I feel _pendingContinuation is already clean enough with the pending prefix and adding might just make it more verbose without any advantage.

nx2/blocks/chat/controllers/chat-controller.js
sharanyavinod: Nice catch, updated.

hannessolo:
Screenshot 2026-06-22 at 14 32 53
Screenshot 2026-06-22 at 14 33 08

As discussed there are still issues with state management.

Suggested appraoch:

- Extract plain UI improvements into separate PR
- Fix state management together with da-agent backend

PR #516, opened 2026-06-17
feat(sdk): add openChatPanel param to setPrompt

hannessolo:
## Summary
- Adds an `openChatPanel` parameter to `setPrompt` in the public SDK (`nx/utils/sdk.js`)
- Defaults to `true` so calling `setPrompt` will open the chat panel by default
- Pass `openChatPanel: false` to set the prompt without touching panel state (backwards-compatible)

## Test plan
- [ ] Call `setPrompt('hello')` from a plugin iframe — chat panel should open and prompt should be set
- [ ] Call `setPrompt('hello', { openChatPanel: false })` — prompt should be set without opening the panel
- [ ] Call `setPrompt('hello', { autoSend: true })` — panel opens and message is sent immediately

### DIFF
--- a/nx/utils/sdk.js
+++ b/nx/utils/sdk.js
@@ -26,8 +26,8 @@ function closeLibrary() {
   port2.postMessage({ action: 'closeLibrary' });
 }
 
-function setPrompt(text, { autoSend = false, openChatPanel = true } = {}) {
-  port2.postMessage({ action: 'setPrompt', details: { text, autoSend, openChatPanel } });
+function setPrompt(text, { autoSend = false } = {}) {
+  port2.postMessage({ action: 'setPrompt', details: { text, autoSend } });
 }
 
 function showPanel(name) {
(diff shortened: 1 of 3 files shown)


nx/utils/sdk.js
```diff
@@ -26,8 +26,8 @@ function closeLibrary() {
   port2.postMessage({ action: 'closeLibrary' });
 }
 
-function setPrompt(text, { autoSend = false } = {}) {
-  port2.postMessage({ action: 'setPrompt', details: { text, autoSend } });
+function setPrompt(text, { autoSend = false, openChatPanel = true } = {}) {
+  port2.postMessage({ action: 'setPrompt', details: { text, autoSend, openChatPanel } });
```
mhaack: Do we need an extra parameter? If somebody sends a prompt to a chat, I think it's safe to assume that one also wants to see it. I think we can always open the chat panel if not open.

mhaack (dismissed):

mhaack (approved):

PR #520, opened 2026-06-18
fix(ew): Adapting UI to closer match mocks

magdalenaorlowska:
fix(ew): Adapting UI to closer match mocks

Issue 1: Background colour should be gray-50
Issue 2: Panels should have a border (.panel-shell & main) - see also https://github.com/adobe/da-live/pull/1016
Issue 3: Resize button should be: full height, transparent when not in use, gray-400 when hovered


hannessolo:
Code for background and resize button looks good, for borders I have the concern [here](https://github.com/adobe/da-live/pull/1016#issuecomment-4841228323) but once we solve that it's also good to go.

sharanyavinod (approved):

PR #522, opened 2026-06-22
fix(chat): fix plan card rendering, task-status persistence, and CSS

anfibiacreativa:
### Summary

- Add nx-campaign-plan-card — renders the agent's multi-step plan for user review with Run / collapse controls
- Add nx-task-item and nx-task-list — per-step running/done status during plan execution
- Add nx-preflight-card — readiness score, progress bar, collapsible categories, per-check pass/fail icons
- Wire all three into renderers.js via :::plan, :::task-item, :::preflight directives and exit_plan_mode / run_preflight tool cards
- Add WTR mocks for nx.js and mdast to enable renderer and component unit testing

### Why

The chat UI needs rich interactive cards to surface the agent's plan and execution progress
Preflight checks give users confidence that generated content meets standards before publishing
Both da-agent and AO backends need parity — the same card must render regardless of which backend is active
What Changed

nx2/blocks/chat/messages/campaign-plan-card.{js,css} — plan card with title, task list, Run button, collapse chevron
nx2/blocks/chat/messages/task-item.{js,css} and task-list.{js,css} — task status indicators
nx2/blocks/chat/messages/preflight-card.{js,css} — readiness score, progress bar, collapsible categories, pass/fail icons
nx2/blocks/chat/constants.js — added all tool names and directive types
nx2/blocks/chat/renderers.js — full wiring: directive path, tool card path, approval card path
web-test-runner.config.mjs + new mocks — nx.js and mdast importmap mocks for testing

### Test Plan

- [ ] Run npm test — all 917 tests pass, 0 failed
- [ ] Open canvas with ref=local, ask agent to create a page — plan card renders, task items update, preflight card appears as final step
- [ ] Confirm Reject / Always approve / Approve buttons appear below preflight card
- [ ] Approve preflight — agent proceeds; Reject — agent stops
- [ ] Test with AO backend — :::preflight directive in streamed text renders the same card
- [ ] Collapse/expand plan card and preflight categories — state persists correctly

### Risks / Follow-ups

Preflight categories depend on what the agent decides to include — no fixed list enforced on the UI side
Image governance for uploads is a separate follow-up
da-nx PR depends on da-agent feat/submit-plan merging first (or landing together)
Fix SITES-46674

Screenshot 2026-06-22 at 12 01 33
Screenshot 2026-06-25 at 09 43 58

Needs to land with https://github.com/adobe-rnd/da-agent/pull/49

Test URLs:
- Before: https://main--frescopa--exp-workspace.hlx.live/
- After: https://feat-plan-tasks--frescopa--exp-workspace.hlx.live/


andreituicu:
Implemented what I discussed with Natalia:
- render the results card based on the governance results (same categories, same checks) -> always what the user configured
- show an approval pop-up after results are displayed, so the agent cannot advance without the user looking at the results (maybe the next step in a skill is publish, the user should have the option to review before publishing, if they wanted to check the content)
- Additional bug fixes that I found along the way which were blocking
  - tool results not being send back when tool required pre-approval
  - tool results not being available on page refresh
  - renditions not being available on refresh

test/loc/glaas/multimodalPageAssets.test.js
```diff
@@ -216,7 +216,7 @@ describe('GLaaS multimodal source preview URL', () => {
 });
 
 describe('GLaaS multimodal image source URLs', () => {
-  it('maps content.da.live to DA Admin /source with the same path', () => {
+  it.skip('maps content.da.live to DA Admin /source with the same path', () => {
```
hannessolo: Is this skipping tests? I don't think we want this.

nx2/blocks/chat/chat-controller.js
```diff
@@ -224,37 +268,57 @@ export default class ChatController {
       this._update();
       if (autoApprove) queueMicrotask(() => this.approveToolCall(toolCallId, true));
       return;
+    } else if (type === AGENT_EVENT.CONTINUATION) {
+      // Post-execution gate: the tool already finished (card is DONE, result shown).
+      // Flag it as awaiting a Continue/Stop decision. Ephemeral (UI-only) — nothing is
+      // pushed to _messages, so a reload simply drops the prompt while the result persists.
```
hannessolo: So if a user reloads the page, they will no longer be prompted to continue? Not sure I have an opinion on whether it's good or bad, just trying to make sure it's intentional.

hannessolo (changes requested):
Please restore the skipped tests.

Also, from what I can see, we're implementing these tool cards in a very hard-coded way locked to our protocol of communication with the agent. IMO that's fine to move fast, after all it's a benefit of controlling our agent. 

But once we move to coworker/CMA, we probably throw the code added here out?

nx2/blocks/chat/renderers.js
```diff
@@ -15,14 +18,120 @@ function toDOM(hast) {
...
+  if (!streamingText || !plan?.tasks?.length) return plan;
+  const updates = buildTaskStatusMap(parseDirectives(streamingText));
+  if (!updates.size) return plan;
+  return {
+    ...plan,
+    tasks: plan.tasks.map((t) => ({ ...t, status: updates.get(t.label) ?? t.status })),
+  };
+}
+
+function mergeTaskItemsIntoPlan(directives) {
```
sharanyavinod: Lets keep this file specific to renderers only and move all such logic ie merge.. into a utils file.

nx2/blocks/chat/chat.js
```diff
@@ -3,10 +3,14 @@ import { loadStyle, hashChange } from '../../utils/utils.js';
...
 import ChatController from './chat-controller.js';
-import { renderMessage, renderApprovalCard } from './renderers.js';
+import { renderMessage, renderApprovalCard, renderContinuationCard } from './renderers.js';
 import './welcome/welcome.js';
 import './prompts/prompts.js';
 import './pills/pills.js';
+import './messages/campaign-plan-card.js';
+import './messages/governance-evaluation-card.js';
+import './messages/task-list.js';
+import './messages/task-item.js';
```
sharanyavinod: chat.js shouldn't have to know about the card internals - we want to limit imports here to only things that are actually necessary for rendering the core chat so the file doesnt grow unnecessarily. Everything else needs to be moved to where it is relevant so this would be within the scope of renderers.

nx2/blocks/chat/messages/governance-evaluation-card.js
```diff
@@ -0,0 +1,340 @@
...
+      return html`<svg class="ge-check-icon ge-check-yes" viewBox="0 0 20 20" aria-hidden="true">
+        <path d="M4 10l4.5 4.5L16 6" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round" fill="none"/>
+      </svg>`;
+    }
+    if (check.alignment === 'NO') {
+      return html`<svg class="ge-check-icon ge-check-no" viewBox="0 0 20 20" aria-hidden="true">
+        <path d="M6 6l8 8M14 6l-8 8" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round" fill="none"/>
+      </svg>`;
+    }
+    return html`<svg class="ge-check-icon ge-check-na" viewBox="0 0 20 20" aria-hidden="true">
```
sharanyavinod: We want to avoid using inline svg, and instead reuse existing s2 icons. See https://github.com/adobe/da-nx/blob/main/nx2/blocks/chat/chat.js#L21 for reference.

nx2/blocks/chat/messages/governance-evaluation-card.js
```diff
@@ -0,0 +1,340 @@
...
+            ${pageUrl ? html`<p class="ge-page-url">${pageUrl}</p>` : nothing}
+            ${this._renderSummaryBar(aggregateSummary)}
+            ${this._renderTextSection(textEvaluation)}
+            ${this._renderImageGroup(imageEvaluations)}
+          </div>
+        ` : nothing}
+      </div>
+    `;
+  }
+}
```
sharanyavinod, line 238: Lets prefer using native elements wherever possible. For example, instead of custom logic for expand/collapse of sections, lets use native details/summary and style that as needed instead. Similarly we have a custom progress bar, where we could instead use native <progress>.

sharanyavinod (changes requested):

test/loc/glaas/multimodalPageAssets.test.js
andreituicu: I see they were skipped in b85d392e, I added them back in 8885700. Not sure why they were removed, they seem unrelated to the changes in this PR and they are passing for me.

nx2/blocks/chat/messages/governance-evaluation-card.js
andreituicu: should be fixed in [3468132](https://github.com/adobe/da-nx/pull/522/commits/346813249486f2bb6333f96824f33ab73fe8ae7e). Thank you!

nx2/blocks/chat/messages/governance-evaluation-card.js
andreituicu: Should be fixed by [5d71144](https://github.com/adobe/da-nx/pull/522/commits/5d7114474530781e932f079756e49b85635e6d3a) and [d23cb45](https://github.com/adobe/da-nx/pull/522/commits/d23cb45018b9028ede41c2c9448d46b878c7ed46). Thank you!

nx2/blocks/chat/renderers.js
andreituicu: Should be fixed by 72f9da11. Thank you!

nx2/blocks/chat/chat.js
andreituicu: Should be fixed by [2328dc9](https://github.com/adobe/da-nx/pull/522/commits/2328dc93dac9c8c469a2854915ce1f41892040bc)! Thanks!

nx2/blocks/chat/chat-controller.js
andreituicu: Yes, that was intentional.
My thinking was that refreshing the page anyway kills the agentic loop, so the user has to write "continue" or something like that to kick if off again, or if they don't want it to continue, they just clear the chat.

I took the example of claude code, where if you are for example in plan mode and it asks whether you want to switch to implementation, if you close claude code and restart it, you don't get the pop-up again, you have to ask it to restart. That looked reasonable to me to replicate.

If you change your mind and would like it differently, happy to adjust!

andreituicu:
@hannessolo 

> Also, from what I can see, we're implementing these tool cards in a very hard-coded way locked to our protocol of communication with the agent. IMO that's fine to move fast, after all it's a benefit of controlling our agent.

> But once we move to coworker/CMA, we probably throw the code added here out?

I think some of the code should be easily portable, as long as Experience Workspace maintains its own chat UI.

For example, the card for rendering the governance results: As long as in the messages returned by the connected harness we can identify the tool calls and tool output, than we can still keep the custom rendering.

The loading state for the tool, will depend on whether the tool invocation and tool results are transmitted to the consumer as independent events, if we just see the finish event or also the start event.

The plan mode and tasks that Natalia implemented, might need to be reconciled with a harness native "Plan Mode" if they provide it, but I've seen the pattern of TODO list vs. full plan mode being implemented in multiple places too, so that might stay as well.

Maybe the part that would be a little complex to keep are the custom approvals, including https://github.com/adobe/da-nx/pull/617 , but that's just something that we'll have to see during porting.

I would like to give it a try, especially for the governance result card before I go to PTO.

So I think that which parts stay and which parts get thrown out really will depend on how the integration between the EW Chat and new harnesses looks like concretly.

andreituicu:
@hannessolo @sharanyavinod 

I resolved the PR feedback. Thank you very much for taking the time to review and for the guidance!
From my point of view, should be ready to merge Monday morning, to have it in production along with: https://github.com/adobe-rnd/da-agent/pull/49

@hannessolo I also merged back and reconciled the main branches with the v2 approvals protocol and re-tested the functionality from the PR.
Everything still works from the point of view of this PR.
Thank you for your plan MD, fed that into claude too.

One part I wasn't sure about, if I should merge the continuation-approvals into the v2 state machine. I don't really have an opinion, personally. Claude said it is better to leave them orthogonal, but if you have an opinion, happy to apply it!

Also if you have any additional feedback, happy to address!

nx2/blocks/chat/messages/campaign-plan-card.js
```diff
@@ -0,0 +1,137 @@
...
+ */
+class NxCampaignPlanCard extends LitElement {
+  static properties = {
+    plan: { attribute: false },
+    _isExpanded: { state: true },
+  };
+
+  constructor() {
+    super();
+    this._isExpanded = true;
```
sharanyavinod: This still doesnt use the native capabilities to expand/collapse - could you take a look please?

nx2/blocks/chat/messages/campaign-plan-card.css
```diff
@@ -0,0 +1,181 @@
...
+}
+
+.plan-header-actions {
+  display: flex;
+  align-items: center;
+  gap: var(--s2-spacing-75);
+  flex-shrink: 0;
+}
+
+.plan-icon-btn {
```
sharanyavinod: There seems to be quite a few reused styles - where possible could we reuse classes from https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html and for the ones which are reused across the cards/list have it within messages css or another shared css file?

nx2/blocks/chat/messages/task-item.js
```diff
@@ -0,0 +1,53 @@
+import { LitElement, html } from 'da-lit';
+import { loadStyle } from '../../../utils/utils.js';
+import { TASK_STATUS } from '../constants.js';
+
+export { TASK_STATUS };
```
sharanyavinod: Why export it here?

nx2/blocks/chat/messages/campaign-plan-card.js
andreituicu: Yes, you are right, I only looked at the governance components. 
Sorry for that! It should now use native elements (summary/details) [fb1a01f](https://github.com/adobe/da-nx/pull/522/commits/fb1a01fa0bb2d8eb8a58d2ddf0b9e719233e342a) and I also found an inlined SVG that I fixed in [2b8d706](https://github.com/adobe/da-nx/pull/522/commits/2b8d706daa5c91ce58beda942ba1b3b4e3bf577d)

nx2/blocks/chat/messages/task-item.js
andreituicu: `TASK_STATUS` was initially defined in this specific JS file and exported for consumers.
In the meantime it was moved in `constants.js`, but the export was not removed.
Should be fixed with [9f3cd1a](https://github.com/adobe/da-nx/pull/522/commits/9f3cd1a0b21061b0a64a6c63181ca9ca2d6a5679), including the removal of additional unused code.

Thank you!

nx2/blocks/chat/messages/campaign-plan-card.css
andreituicu: With [89e2f48](https://github.com/adobe/da-nx/pull/522/commits/89e2f488d7eae01ace83384880e1aec1bc5a67b6) I reused the nx buttons and refactored the common CSS for the cards in: `nx2/blocks/chat/messages/messages.css`

I think the part that could be discussed is the spinner, if you'd like to reuse the one from NX, the agent's analysis was that it would be a bit too much to load `forms.css` just for the purpose of the spinner and I understand the argument, but I'm happy to adjust if you prefer that option.

anfibiacreativa:
Superseeded ny #658

PR #548, opened 2026-06-24
updated svg util from Loc PR so that it can be used for other blocks migration

ravuthu:

Fix #<gh-issue-id>

Test URLs:

### DIFF
--- a/nx2/utils/svg.js
+++ b/nx2/utils/svg.js
@@ -20,6 +20,50 @@ export const loadHrefSvg = (() => {
   };
 })();
 
+const iconCache = new Map();
+
+async function fetchIcon(path) {
+  // Check if we already have a request in flight or completed
+  if (iconCache.has(path)) {
+    const cachedSvg = await iconCache.get(path);
+    // Clone the node because an element can only exist in one place in the DOM
+    return cachedSvg ? cachedSvg.cloneNode(true) : null;
+  }
+
+  const fetchPromise = (async () => {
+    try {
+      const resp = await fetch(path);
+      if (!resp.ok) return null;
+
+      const text = await resp.text();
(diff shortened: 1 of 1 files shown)


nx2/utils/svg.js
```diff
@@ -20,6 +20,50 @@ export const loadHrefSvg = (() => {
   };
 })();
 
+const iconCache = new Map();
+
+async function fetchIcon(path) {
```
anfibiacreativa: isn't this exactly what loadHrefSvg above already does? we're duplicating fetch+cache in the same file. why not extend loadHrefSvg to support cloning instead of a second cache?

nx2/utils/svg.js
```diff
@@ -20,6 +20,50 @@ export const loadHrefSvg = (() => {
...
+  // Check if we already have a request in flight or completed
+  if (iconCache.has(path)) {
+    const cachedSvg = await iconCache.get(path);
+    // Clone the node because an element can only exist in one place in the DOM
+    return cachedSvg ? cachedSvg.cloneNode(true) : null;
```
anfibiacreativa: we cache the promise even when it resolves to null on failure. a transient network blip permanently breaks that icon until reload. please cache successful results only and let failures retry

nx2/utils/svg.js
```diff
@@ -20,6 +20,50 @@ export const loadHrefSvg = (() => {
   };
 })();
 
+const iconCache = new Map();
```
anfibiacreativa: iconCache never gets bounded. probably fine for a fixed icon set, but worth a size cap or comment if paths can ever be dynamic

anfibiacreativa (changes requested):

PR #554, opened 2026-06-25
style: improve breadcrumb positioning

sharanyavinod:
Test URLs:

Previously:
https://ewbwse--da-live--adobe.aem.live/#/exp-workspace/frescopa

Current:
https://da.live/canvas?nx=ew-crmb#/exp-workspace/frescopa/index
https://ewbwse--da-live--adobe.aem.live/?nx=ew-crmb#/exp-workspace/frescopa

### DIFF
--- a/nx2/blocks/chat/chat.css
+++ b/nx2/blocks/chat/chat.css
@@ -149,7 +149,7 @@
   width: 10px;
   height: 10px;
   background-color: currentcolor;
-  mask-image: url("/nx2/img/icons/S2_Icon_ChevronUp_20_N.svg");
+  mask-image: url("https://da.live/img/icons/s2-icon-chevronup-20-n.svg");
   mask-size: contain;
   mask-repeat: no-repeat;
   mask-position: center;
--- a/nx2/blocks/chat/chat.css
+++ b/nx2/blocks/chat/chat.css
@@ -149,7 +149,7 @@
   width: 10px;
   height: 10px;
   background-color: currentcolor;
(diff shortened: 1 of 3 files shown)


nx2/blocks/chat/chat.css
```diff
@@ -149,7 +149,7 @@
   width: 10px;
   height: 10px;
   background-color: currentcolor;
-  mask-image: url("/nx2/img/icons/S2_Icon_ChevronUp_20_N.svg");
+  mask-image: url("https://da.live/img/icons/s2-icon-chevronup-20-n.svg");
```
hannessolo: Hm, not sure we should be hardcoding da.live on da-nx...

Will this load if we're on a different domain?

sharanyavinod: Good point, I cleaned it up and other usage on da-nx to be relative. 
How do you foresee usage on a different domain? For decorative icons I really want to avoid loading the svg from the js - would rather even keep the couple or so such icons on da-nx instead for such cases. Wdyt?

hannessolo: Well, "different domain" is very hypothetical but wouldn't it already break on da-live on aem.page?

mhaack (approved):

nx2/blocks/chat/chat.css
sharanyavinod: Switched to relative, and the icons are also on preview, so should be good

PR #555, opened 2026-06-25
fix: make nx2 crawl backend-aware via source.list

chrischrischris:
## Problem

The crawler in `nx2/public/utils/tree.js` (`getChildren`) listed folder contents by hitting `${DA_ORIGIN}/list${path}` directly, bypassing the backend-aware `source.list`. On a **Helix 6** site that endpoint has nothing, so any crawl returns no files.

This surfaced in da-live as [adobe/da-live#1034](https://github.com/adobe/da-live/issues/1034): deleting a folder crawls it to count its files before showing the confirmation dialog, which reported **"0 items"** and broke the flow on hlx6 sites.

## Fix

Route `getChildren` through `source.list` (from `nx2/utils/api.js`), which:
- detects the backend per-site via `isHlx6` and hits the right origin (AEM API for hlx6, DA `/list` for legacy), and
- returns DA-normalized items (`{ name, path, ext, ... }`) — the exact shape `getChildren` already separates into files vs folders.

Removed the now-unused `daFetch` / `DA_ORIGIN` imports.

**Compatibility:** Legacy DA behavior is unchanged — DA *is* the backend, and `source.list` falls back to the same `/list` endpoint (`DA_ORIGIN` and `DA_ADMIN` resolve to the same origin). Only Helix 6 sites — previously broken — change. nx1's `tree.js` is intentionally left DA-only (nx1 is the legacy DA-only world). The only consumer of nx2's crawl is da-live; there are no internal nx2 callers.

## Tests

Added `test/nx2/public/utils/tree.test.js`:
- crawls a Helix 6 site via `source.list` (the #1034 regression)
- crawls the legacy DA backend via the `source.list` fallback
- follows the continuation token across pages
- stops a folder listing on a non-ok response

Full suite: **935 passing**. (Note: the pre-existing nx1 `tree.test.js` "Respects throttle parameter" test is timing-flaky under concurrent load — `setTimeout(50)` occasionally measures 49ms — and is unrelated to this change.)

### DIFF
--- a/WORKLOG.md
+++ b/WORKLOG.md
@@ -1,5 +1,15 @@
 # Worklog
 
+## 2026-06-25
+
+### nx2/public/utils/tree.js — backend-aware crawl (crawlhlx6 branch)
+
+`getChildren` listed folder contents by fetching `${DA_ORIGIN}/list${path}` directly, bypassing the backend-aware `source.list`. On a Helix 6 site that endpoint has nothing, so da-live's delete confirmation — which crawls a folder to count its files before showing the dialog — reported "0 items" and the delete flow broke (adobe/da-live#1034).
+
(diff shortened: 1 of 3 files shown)


mhaack:
@chrischrischris is this the same as https://github.com/adobe/da-nx/pull/553?

mhaack (approved):

mhaack:
@chrischrischris is this still needed? Looks like it has only tests after merging latest main.

PR #562, opened 2026-06-26
feat(ew): render layout-mode comment markers in the quick-edit iframe

usman-khalid:
To be merged with https://github.com/adobe/da-live/pull/1073

### DIFF
--- a/nx/public/plugins/quick-edit/src/comments.js
+++ b/nx/public/plugins/quick-edit/src/comments.js
@@ -1,0 +1,529 @@
+import { MESSAGE_TYPES } from '../../../../utils/message-types.js';
+import { parseIndex, positionBox } from './utils.js';
+import {
+  findTextBlock,
+  findBlock,
+  findImageAtProseIndex,
+  pictureSrc,
+  srcPathsMatch,
+  safeQuerySelectorAll,
+  walkProsePositions,
+} from './dom-index.js';
+
+const OVERLAY_ID = 'qe-comment-overlay';
+
+function blockClassFromAnchor(anchorText) {
+  return String(anchorText || '').replace(/^block:\s*/, '').trim().split(/\s+/)[0] || '';
+}
+
+// The anchored block node spans the PM range [from, to). getInstrumentedHTML
+// stamps data-block-index = posAtDOM(node, 0), i.e. the node's content-start
+// (from + 1), which is the unique block index inside that range. Selecting the
+// smallest in-range index returns the outermost (anchored) block when blocks
+// are nested. Because the range comes from a Yjs relative position, this is an
+// exact, per-instance match — duplicate block types are never confused.
+function findBlockByRange(from, to, root) {
+  const rangeEnd = typeof to === 'number' && to > from ? to : from + 1;
+  let best = null;
+  let bestIndex = Infinity;
+  root.querySelectorAll('[data-block-index]').forEach((el) => {
+    const idx = parseIndex(el.getAttribute('data-block-index'));
+    if (idx == null || idx < from || idx >= rangeEnd) return;
+    if (idx < bestIndex) {
+      bestIndex = idx;
+      best = el;
+    }
+  });
+  return best;
+}
+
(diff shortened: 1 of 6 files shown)


nx/utils/message-types.js
```diff
@@ -25,6 +27,9 @@ export const MESSAGE_TYPES = Object.freeze({
   NEW_VERSION: 'new-version', // no payload
   SELECTION_CHANGE: 'selection-change', // { anchor, head, anchorX, anchorY }
   STORED_MARKS: 'stored-marks', // { marks: PM Mark.toJSON()[] }
+  COMMENT_MARKER_CLICK: 'comment-marker-click', // { payload: { threadId: string } }
+  COMMENT_MARKER_CLEAR: 'comment-marker-clear', // no payload
+  COMMENT_SHORTCUT: 'comment-shortcut', // no payload
```
sharanyavinod, line 32: For all new message types introduced, please update the relevant doc with the details around it.

nx/public/plugins/quick-edit/src/comments.js
```diff
@@ -0,0 +1,490 @@
...
+// The initials bubble replaces the plain dot for every anchor type: a small
+// teardrop chip tinted with the author's color, showing their initials.
+function createMarkerBubble({
+  threadId, isActive, rect, overlay, ctx, placement = 'top-left', color, textColor, initials,
+}) {
+  const bubble = document.createElement('button');
+  bubble.type = 'button';
+  bubble.className = `qe-comment-bubble${isActive ? ' is-active' : ''}`;
+  bubble.dataset.commentThread = threadId;
+  bubble.setAttribute('aria-label', 'Open comment');
```
sharanyavinod: Even though aria-label is set here, the parent overlay has aria-hidden set to true, which removes all its children from the accessibility tree.

nx/public/plugins/quick-edit/src/comments.js
```diff
@@ -0,0 +1,490 @@
...
+function setupCommentClicks(ctx) {
+  if (ctx.commentClickListener) return;
+  ctx.commentClickListener = (event) => {
+    const threadId = textHighlightThreadAtPoint(event.clientX, event.clientY);
+    if (threadId) {
+      ctx.port?.postMessage({ type: MESSAGE_TYPES.COMMENT_MARKER_CLICK, payload: { threadId } });
+      return;
+    }
+    if (!ctx.selectedThreadId) return;
+    if (event.target.closest('.qe-comment-box, .qe-comment-marker, .qe-comment-bubble')) return;
```
sharanyavinod: Where is qe-comment-marker used?

nx/public/plugins/quick-edit/src/comments.js
```diff
@@ -0,0 +1,490 @@
...
+export function setupCommentShortcut(ctx) {
+  if (ctx.commentShortcutListener) return;
+  ctx.commentShortcutListener = (event) => {
+    const isShortcut = (event.metaKey || event.ctrlKey) && event.altKey && event.code === 'KeyM';
+    if (!isShortcut) return;
+    event.preventDefault();
+    ctx.port?.postMessage({ type: MESSAGE_TYPES.COMMENT_SHORTCUT });
+  };
+  document.addEventListener('keydown', ctx.commentShortcutListener);
+}
```
sharanyavinod: I think this file is a bit too huge currently mixing a lot of concerns. Might be nice if it could be split into meaningful pieces for eg. rendering, utils, orchestration etc. Wdyt?

nx/public/plugins/quick-edit/src/comments.js
usman-khalid: It wasn't. :) thanks, removed.

nx/public/plugins/quick-edit/src/comments.js
usman-khalid: Agree, moved to a `comments` with sparation of concerns.

sharanyavinod (dismissed):

mhaack (approved):

PR #568, opened 2026-06-29
fix: extend breadcrumb

sharanyavinod:
Extend breadcrumb to support actions at leaf level, suffix elements, and custom font size

Preview: https://da.live/canvas#/exp-workspace/frescopa/index


hannessolo:
Where can I see the suffix? I don't understand where it's being used

sharanyavinod:
> Where can I see the suffix? I don't understand where it's being used

Just added a slot at the end, so we can include things if needed - like a badge/extra separator etc. For eg. in browse there is an additional separator and button which we stitch together, but for things that are semantically and visually supposed to be part of the breadcrumb, makes sense its included here. Wdyt?

hannessolo:
Ok, but we don't have a use case for it yet?

I think generally we should add stuff as we need it, but since you already wrote it let's keep it.

hannessolo (approved):

sharanyavinod:
> Ok, but we don't have a use case for it yet?
> 

We do - gear action button, separator after breadcrumb etc are already on the browse breadcrumb. These are currently absolutely positioned to look like they belong together.

PR #571, opened 2026-06-30
fix(hlx6): Filter hidden files and folders from hlx6 list results

### LINKED ISSUE #1053: [hlx6] .trash folder is visible in the browse view
Image

mhaack:
DA admin does not return hidden folders like `.trash` or `.da` in list responses, but hlx6 does. This aligns the behavior by filtering any item whose name starts with `.` in `hlx6ToDaList` before normalizing the results.

fix: https://github.com/adobe/da-live/issues/1053

- Added filter in `hlx6ToDaList` to exclude hidden files/folders
- Added test: `source.list hlx6 filters out hidden files and folders starting with "."`

Test:
- https://da.live/?nx=hfold#/da-testautomation/da-e2e-tests


nx2/utils/api.js
```diff
@@ -629,7 +629,7 @@ async function callPath({
 }
 
 function hlx6ToDaList(parentPath, items) {
-  return items.map((item) => {
+  return items.filter((item) => !item.name.startsWith('.')).map((item) => {
```
kptdobe, line 632: Can `item.name` be null ? Sounds stupid but I have seen documents with empty names in the list... did not find how they were created but they exist.

kptdobe (approved):

nx2/utils/api.js
mhaack: Good question, saw these as well. Not sure if that can happen with HLX6 backend? @bosschaert do you know?

sharanyavinod (approved):

nx2/utils/api.js
bosschaert: It can definitely happen on da-admin. See in this screenshot for an example, where a subfolder shows up with its name being an empty string:

Image

I don't think it can happen with Helix 6, but maybe be on the safe side and check for it anyway.

mhaack: Created https://github.com/adobe/da-nx/pull/573 as follow up

PR #572, opened 2026-07-01
feat: add skill scripts runtime

anfibiacreativa:
THIS CODE IS NOT MEANT TO STAY PERMANENTLY. IT'S AN ITERIM SOLUTION FOR CODE EXECUTION.

## Summary

- Adds a worker substrate that executes skill scripts triggered by the da-agent orchestration tool (PR 1).
- Wires up marketplace resolution + `host.deps` injection so scripts can call DA APIs without bundling credentials.
- Converts `attachmentRef` to bytes for binary-capable skill steps.

## Why

Skills need a sandboxed execution environment inside da-nx so scripts fetched from the marketplace (see da-agent PR 1) can run safely and interact with the document. **This PR is co-dependent with da-agent PR 1 — ship them together.**

The demo marketplace source (`exp-workspace/skills`, `ew/` tree) is hardcoded for now and should move to the config sheet (targeting `adobe/skills`) for prod. See the skill-script-runtime migration plan.

## What Changed

- Worker substrate: sandboxed script execution (localStorage neutered, document access gated)
- Agent-triggered round-trip: da-agent `skill_run_script` → da-nx worker → result back to agent
- Marketplace resolution: worker resolves skill body via provider URL, caches per session
- `host.deps` injection: injects authenticated DA API helpers into skill script scope
- `attachmentRef` → bytes: binary payload conversion for attachment-capable skills
- Chat empty-directive render fix: blank directive no longer renders as broken UI element
- `.docx` accept: file picker now accepts `.docx` uploads
- Design docs: worker substrate architecture and sandbox isolation notes

## Test Plan

- [ ] `npm test` passes (all relevant test suites)
- [ ] Security suite: sandbox isolation — localStorage inaccessible from script scope; document write blocked outside allowed surface; no credential bleed into script scope
- [ ] Agent round-trip: send a chat message that invokes a marketplace skill end-to-end; script executes and returns structured result
- [ ] `host.deps` injection: skill script can call a DA API helper without holding its own credentials
- [ ] `attachmentRef` → bytes: upload a file attachment; skill receives binary payload correctly
- [ ] Chat empty-directive fix: send a message with no directive; no broken UI element appears
- [ ] `.docx` accept: file picker shows `.docx` as an accepted type

## Risks / Follow-ups

- Demo marketplace source `exp-workspace/skills` (`ew/`) is hardcoded — must move to config sheet targeting `adobe/skills` before prod
- Co-dependent with da-agent PR 1; do not ship independently
- Sandbox isolation is enforced in the worker; review surface area before widening `host.deps`

---

## Reference

**PR 2 of 4 — skill-script marketplace + runtime.**
Co-dependent pair: **da-agent (PR 1) ↔ da-nx (PR 2) — ship together.**
ew-extensions (PR 3) and ew-devs-cli (PR 4) are standalone.

## Test URLs

| | URL |
|---|---|
| Before | https://main--da-nx--adobe.aem.page/ |
| After | https://feat-skill-scripts-runtime--da-nx--adobe.aem.page/ |


nx2/utils/skill-runtime/worker-host.js
```diff
@@ -0,0 +1,65 @@
+// Dependency allowlist: name → vetted module URL served by this host.
+// Skills declare deps in execution_dependencies; only names in this map are permitted.
+// A skill declaring any name NOT present here is refused before execution.
+export const DEPENDENCY_ALLOWLIST = {
+  fflate: '/nx2/deps/fflate/dist/index.js',
```
hannessolo, line 5: IIUC this is something we control - so how will we decide which dependencies we provide?

docs/skill-script-runtime.md
```diff
@@ -0,0 +1,513 @@
...
+### 2.2 Execution metadata
+
+Authored skills declare an `execution` block in `skill.md` frontmatter; built-in
+(in-repo) proof skills declare the equivalent as a plain manifest object:
+
+```yaml
+execution:
+  entry: convert          # exported function name
+  runtimes: [js]          # implementations present (js | py)
+  capabilities: []        # [] => pure compute => client-eligible
```
hannessolo, line 71: This is the first time capabilities are mentioned - perhaps we should already here either document which are available, or link to the `capablities.js` file where they are defined.

docs/skill-script-runtime.md
```diff
@@ -0,0 +1,513 @@
...
+- a pure script runs in the worker and returns serializable output;
+- `fetch` (and the other network/storage globals) is `undefined` inside the worker —
+  enforce-by-construction holds;
+- a manifest with `capabilities: ['network']` returns `{ error: 'requires server runtime' }`
+  **without** spinning a worker;
+- a runaway script is killed at `timeoutMs`;
+- the docx skill converts a fixture (`<w:t>hello world</w:t>` → markdown), unescapes
+  XML entities, and returns `{ error }` on corrupt input without throwing past the runner.
+
+**Explicitly out of scope this round:** wiring into the chat attachment flow (a
```
hannessolo, line 168: Is hooking into the chat attachment be the only non-agent trigger that we have planed? Should we already define how this hook will look like in the skill?

nx2/utils/skill-runtime/worker-host.js
anfibiacreativa: I am thinking that your question comes from the assumption that customers will load their own skills with scripts and how will we know what they need, in order to provide it?
Which is a very valid question in that context. But while we don't have a server-side harness, we would only use this capability for our own skills and hence decide what we need to provide, ourselves.

mhaack:
Since tis applies to the old chat/agent I think we can close it for now.

PR #573, opened 2026-07-01
fix(hlx6): guard hlx6ToDaList against empty or missing item names

mhaack:
Follow-up to #571. `!item.name.startsWith('.')` would throw on `null`/`undefined` and silently pass through empty strings.

The filter is scoped to hlx6 items only (those with a `content-type`): legacy DA items still pass through untouched so that `tree.js` can handle its own empty-name logging as before.

- Updated filter: only applies when `item['content-type']` is present
- Added test: `source.list hlx6 filters out items with empty or missing name` (covers `''`, `null`, and missing `name` property)
Test:

https://da.live/?nx=hfold#/da-testautomation/da-e2e-tests


test/nx2/utils/api.test.js
```diff
@@ -273,6 +273,23 @@ describe('api.js', () => {
...
     });
 
+    it('source.list hlx6 filters out items with empty or missing name', async () => {
+      restoreFetch();
+      installFetch({
+        body: JSON.stringify([
+          { name: 'visible.html', 'content-type': 'text/html' },
+          { name: '', 'content-type': 'application/folder' },
+          { name: null, 'content-type': 'application/folder' },
+          { 'content-type': 'application/folder' },
```
kptdobe, line 284: No test for file with empty name ?

kptdobe (dismissed):

test/nx2/utils/api.test.js
mhaack: Added a case with a file (non-folder, e.g. text/html) having an empty name in the same test — previously only folder items covered the empty/missing-name path.

mhaack:
@kptdobe can you please 👍 again, I added the missing test

kptdobe (approved):

PR #577, opened 2026-07-06
feat: implement feedback

mhaack:
Implement the feedback functionality with the dialog for the end-user to submit ideas and bug requests for DA.
 
The backend worker is in https://github.com/adobe-rnd/da-feedback and deployed to feedback.da.live. It takes care of validating the messages and forwarding them to the Slack channel. 

The entries for the feedback menu are just content in /fragments/nav/feedback page in the da.live site content

Test URLs:
- Test: https://da.live/drafts/mhaack/browse?nx=feedback#/exp-workspace/frescopa


nx2/blocks/feedback/feedback-dialog.js
```diff
@@ -0,0 +1,189 @@
...
+const KIND_LABELS = { idea: 'Idea', bug: 'Bug' };
+
+/**
+ * Feedback dialog: a title + category dropdown + free-text textarea + a
+ * "link my current chat session" checkbox + Cancel/Submit actions.
+ */
+class NxFeedbackDialog extends LitElement {
+  static properties = {
+    label: { attribute: false },
+    kind: { attribute: false },
```
sharanyavinod: This might not need to be a reactive property

sharanyavinod:
Is the uploaded svg required?

nx2/styles/form.css
```diff
@@ -0,0 +1,149 @@
+.da-input {
```
sharanyavinod, line 1: Can we have these styles outside of dialog? Since theoretically they could be used in any context, not necessarily just within a dialog. Wdyt?

mhaack:
> Is the uploaded svg required?

Yes as the nav menu itself use the old pattern. To not break anything I prefere to keep if like this for now.

nx2/blocks/feedback/feedback-dialog.js
mhaack: Removed. `kind` is set once via `dialog.kind = item.id` before the element connects and is only read inside `_handleSubmit`, never referenced in `render()`, so it doesn't need to be a reactive/tracked property. Fixed in 3be7d25.

nx2/styles/form.css
mhaack: Moved to nx2/styles/form.css and updated the imports.

sharanyavinod (dismissed):

sharanyavinod (approved):

PR #578, opened 2026-07-06
feat(profile): move Legal notices entry to profile menu (to remove the dedicated help icon)

mhaack:
## Summary

Adds a "Legal notices" entry to the profile popover menu that opens the shared `/fragments/nav/help` content in a dialog.

## Changes

- **`nx2/blocks/profile/profile.js`**: added `LEGAL_NOTICES_PATH` (`/fragments/nav/help`), a local `openFragmentDialog()` helper (mirrors `nav.js`), and `handleLegalNotices()` wired to a new "Legal notices" menu item in the Links section.
- **`nx2/blocks/profile/profile.css`**: styled the new button to match the existing menu links.
- **`nx2/test/mocks/ims.js`**: extended the shared IMS mock with `getIo()`/`getOrgs()` (plus `userId`/`displayName`/`email`) support so signed-in component tests can render past the loading state. Purely additive — the one existing consumer (`utils.test.js`) is unaffected.
- **`nx2/test/unit/nx/blocks/profile/profile.test.js`** (new): first test coverage for `nx-profile`, covering the anonymous sign-in render, the new "Legal notices" menu entry, and the fragment dialog opening on click (fetch mocked, following the `nav.test.js`/`feedback.test.js` conventions).

Test:
- https://da.live/drafts/mhaack/browse?nx=legal#/exp-workspace/frescopa


nx2/blocks/profile/profile.js
```diff
@@ -6,7 +6,23 @@ import { signout } from '../../utils/api.js';
 
 const config = getConfig();
 
+// Well-known, hardcoded path (not configurable) for the legal notices
+const LEGAL_NOTICES_PATH = '/fragments/nav/help';
+
 const style = await loadStyle(import.meta.url);
+
+async function openFragmentDialog(path) {
```
mhaack, line 21: @hannessolo this was copied from `nav.js` which was used to open the dialog before. In `nav.js`, it's actually more generic and can open any dialog if the nav fragment page contains a fragment link. However, it was only used for this one dialogue, as far as I know. So maybe it should be removed from `nav.js` to not leave tech dept. WDYT?

sharanyavinod:
Couple questions: 
- Quite a bit of space around the content, maybe we can reduce it a bit?
- Since this dialog has no action buttons, we might need an explicit close option at the top?

nx2/blocks/profile/profile.js
anfibiacreativa: openFragmentDialog is async and can reject (dynamic import or loadFragment throwing), and we're firing it and forgetting. can we wrap the body of openFragmentDialog in a try/catch so a failed fragment fetch doesn't leak an unhandled rejection?

nx2/blocks/profile/profile.js
```diff
@@ -82,6 +98,10 @@ class NxProfile extends LitElement {
     this.dispatchEvent(event);
   }
 
+  handleLegalNotices() {
+    openFragmentDialog(LEGAL_NOTICES_PATH);
```
anfibiacreativa, line 124: NIT: our handler convention is _onLegalNotices. i know handleScheme/handleSignOut already break this, so either we match the new one to those or fix them all, but let's not add a third spelling. worth a follow-up to standardise?

nx2/test/unit/nx/blocks/profile/profile.test.js
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  it('opens a fragment dialog with the shared legal notices content when clicked', async () => {
+    restoreFetch = mockFragmentFetch();
+    const el = await createProfile();
+    await waitForSignedIn(el);
+
+    const legalBtn = el.shadowRoot.querySelector('.nx-menu-links .nx-menu-link-btn');
+    legalBtn.click();
+    await new Promise((r) => { setTimeout(r, 50); });
```
anfibiacreativa: why the fixed 50ms sleep? we already poll in waitForSignedIn, can we poll for document.querySelector('nx-dialog') the same way instead of racing a timer?

nx2/test/unit/nx/blocks/profile/profile.test.js
```diff
@@ -0,0 +1,90 @@
...
+  it('renders a "Legal notices" entry in the profile menu when signed in', async () => {
+    const el = await createProfile();
+    await waitForSignedIn(el);
+
+    const legalBtn = el.shadowRoot.querySelector('.nx-menu-links .nx-menu-link-btn');
+    expect(legalBtn).to.not.be.null;
+    expect(legalBtn.textContent.trim()).to.equal('Legal notices');
+  });
+
+  it('opens a fragment dialog with the shared legal notices content when clicked', async () => {
```
anfibiacreativa, line 101: we test the happy path but not the null-fragment branch. can we add a case where the fragment fetch fails/returns nothing and assert no nx-dialog gets appended?

anfibiacreativa (changes requested):

nx2/blocks/profile/profile.js
mhaack: Fixed. Wrapped the body in try/catch (mirrors the log-and-swallow pattern already used elsewhere in this file). 14715e6.

nx2/blocks/profile/profile.js
mhaack: Checked this file - every existing handler here uses handle* (handleLoaded, handleCopyUser, handleOrgSwitch, handleSignOut, handleScheme), none use _on*. Renaming just handleLegalNotices would make it the only outlier in this file, so keeping it as-is for local consistency. Agree a standardization pass across the codebase is worth a follow-up ticket rather than a per-PR fix.

nx2/test/unit/nx/blocks/profile/profile.test.js
mhaack: Fixed. Added a generic waitFor() poller (same pattern as waitForSignedIn) and used it instead of the fixed sleep. 14715e6.

nx2/test/unit/nx/blocks/profile/profile.test.js
mhaack: Added a test that mocks a 404 fragment fetch and asserts no nx-dialog is appended. 14715e6.

mhaack:
@sharanyavinod @anfibiacreativa please have a second look.

Addressed the feedback above on the code.

Also tightened the dialog panel padding and added a Close action button. Used a close button to stay in line with other dialogs, no X in the top corner.

sharanyavinod (approved):

anfibiacreativa (approved):

PR #580, opened 2026-07-07
feat(ew): move chat's AI disclaimer outside its bordered body

magdalenaorlowska:
Adds an optional `styled` property to nx-chat, reflected as the `styled-as-panel` attribute. When set, nx-chat wraps its header, messages, and input in a bordered/rounded/backgrounded `.chat-body`, and renders the AI disclaimer below it, outside that border — all within its own shadow DOM. styles.css removes the outer panel's own background, border, and border-radius whenever it contains an element with the `styled-as-panel` attribute, so the two don't stack. panel.js is unchanged: it has no knowledge of chat or any specific consumer, just a generic `:has([styled-as-panel])` check.

### DIFF
--- a/nx2/blocks/chat/chat.css
+++ b/nx2/blocks/chat/chat.css
@@ -76,6 +76,29 @@
   }
 }
 
+:host([styled-as-panel]) {
+  padding-top: 0;
+}
+
+.chat-body {
+  display: flex;
+  flex-direction: column;
+  flex: 1 1 auto;
+  min-height: 0;
+  width: 100%;
+  align-items: center;
+  gap: var(--s2-spacing-300);
+  box-sizing: border-box;
+}
+
+:host([styled-as-panel]) .chat-body {
+  border: 1px solid var(--s2-gray-100);
+  border-radius: var(--s2-corner-radius-800);
+  background: light-dark(#fff, #000);
+  overflow: hidden;
+  padding: var(--s2-spacing-200) 0;
+}
+
 .chat-header {
   display: flex;
   align-items: center;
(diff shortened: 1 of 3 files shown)


nx2/blocks/chat/chat.js
```diff
@@ -33,6 +33,7 @@ class NxChat extends LitElement {
     thinking: { type: Boolean },
     connected: { type: Boolean },
     toolCards: { type: Object },
+    styled: { type: Boolean, reflect: true, attribute: 'styled-as-panel' },
```
anfibiacreativa, line 36: styled is a boolean, we prefer to prefix those with is/has, conventionally. the prop/attr mismatch is also confusing. why not isPanel: { type: Boolean, reflect: true, attribute: 'styled-as-panel' } so the intent is clear on both sides?

nx2/blocks/chat/chat.css
```diff
@@ -76,6 +76,29 @@
...
+  width: 100%;
+  align-items: center;
+  gap: var(--s2-spacing-300);
+  box-sizing: border-box;
+}
+
+:host([styled-as-panel]) .chat-body {
+  border: 1px solid var(--s2-gray-100);
+  border-radius: var(--s2-corner-radius-800);
+  background: light-dark(#fff, #000);
```
anfibiacreativa, line 97: we use tokens for colors everywhere else in this file. is there a reason not to use light-dark(var(--s2-gray-50), var(--s2-gray-900)) (or whatever the white/black tokens are) here?

sharanyavinod (changes requested):
Tbh I dont agree with this approach - styling is a pure ux concern and this shouldnt be creeping into the core logic nor requiring a reactive property imo. Should be either controlled by classes or css properties. The panel can then check for panel-wrapper:has(nx-chat) or something similar and we only touch the css. Why not go in this direction?

hannessolo:
To add to @sharanyavinod 's comment above - I'm also not sure we should be doing this. We built the panels system to be generic but in the end if it's only being used in 2 places so far - the chat and the right panel. If one of those places already becomes an exception, we should be wondering if that generic approach makes sense at all or if we need to build it in a different way. 

So I guess my question is:
1) Do we really need this? I agree it looks better but wonder if it's worth the effort.
2) If we do really need this, are we ok with this approach or do we restructure the panels?

PR #333, opened 2026-04-09
feat: preserve conversation history

sharanyavinod:
- Conversation history is now preserved per room: IndexedDB (da-chat DB), keyed by {org}--{site}--{userId}
- Added connect() with exponential-backoff retry (up to 30s cap) before loading history: input disabled until agent is reachable
- "Clear chat" button appears in the header when there are messages

Test URLs:
https://main--da-live--adobe.aem.page/drafts/hhertach/chat?nx=ew-store-chat&nxver=2&da-admin=stage#

Note: Ignore the dummy prompts, added now for easier preview until we integrate in the frame
Note 2: Clear style is draft for now, will update later if needed.

PR #581, opened 2026-07-07
Media lib Quality Improvements

kmurugulla:

Fix #<gh-issue-id>

Test URLs:
- Before: https://main--da-live--adobe.aem.live/apps/media-library#/kmurugulla/brightpath
- After: https://main--da-live--adobe.aem.live/apps/media-library?nx=medialib-quality#/kmurugulla/brightpath

Notes 
  - Replace hardcoded HTTP status codes (401, 403, 500) with HttpStatus constants across 12 files
  - Centralize CSS background colors into custom properties (--ml-bg, --ml-surface, --ml-topbar-bg) for consistent theming
  - Fix error propagation in worker to preserve status codes without fragile string parsing
  - Add error hints to permission-denied errors for better user guidance
  - Centralize hardcoded user messages into core/messages.js for maintainability
  - Add user-friendly validation error messages with actionable suggestions
  - Fix dark mode backgrounds to use consistent color system
  - Clarify medialog access requirements in error messages (author role or higher needed)
  - Fix plugin mode preview authentication using daFetch.initIms


nx/blocks/media-library/core/paths.js
```diff
@@ -212,7 +212,7 @@ export async function validateSitePath(sitePath) {
         return { valid: false, error: t('VALIDATION_SITE_NOT_FOUND', { path: `/${org}/${repo}` }) };
       }
 
-      if (resp.status === 401 || resp.status === 403) {
+      if (resp.status === HttpStatus.UNAUTHORIZED || resp.status === HttpStatus.FORBIDDEN) {
```
mhaack: Not sure if this is really needed? The status numbers are pretty common, don't know if we should have a constant for this.

nx/blocks/media-library/media-library.css
```diff
@@ -7,6 +7,10 @@
   overflow: hidden;
   color-scheme: light dark;
 
+  --ml-bg: light-dark(rgb(249 249 249), rgb(28 28 28));
+  --ml-surface: light-dark(rgb(255 255 255), rgb(26 26 26));
+  --ml-topbar-bg: light-dark(rgb(248 248 248), rgb(27 27 27));
```
mhaack: IHMO we should stick with the spectrum styles. With NX2 you will also get light/dark mode

nx/blocks/media-library/ui/data.js
mhaack: `t` is pretty short. I think is this on other places as well.

nx/blocks/media-library/core/utils.js
```diff
@@ -145,8 +146,8 @@ export async function checkSiteAuthRequired(org, repo) {
 
 export async function livePreviewLogin(owner, repo) {
   try {
-    const { loadIms } = await import('../../../../nx2/utils/ims.js');
-    const { accessToken } = await loadIms() || {};
+    const { initIms } = await import('../../../utils/daFetch.js');
```
mhaack: While this might work we should use NX2, so we can remove NX.

Also there is `const { loadIms } = await import('../../utils/ims.js');`util which is maybe used better here https://github.com/adobe/da-nx/blob/main/nx/utils/ims.js

kmurugulla: We're using nx here instead of nx2 because nx provides the `setImsDetails(token)` function in `daFetch.js` which enables token injection in plugin mode. This capability is critical for the SDK integration where tokens need to be injected externally.

Currently, nx2 doesn't have this token injection capability. Should we enhance nx2 to support this use case so we can eventually migrate away from nx?

mhaack: Not sure I fully understand the flow and usage of `setImsDetails(...)` are you doing a different IMS login?

Also would like to get a thought from @andreituicu on this from security pov.

kmurugulla: Good question! Here's the flow:

## Current Implementation (using nx)

The media library supports **plugin mode** where it runs embedded in another application (like Adobe Express or other Adobe tools). In this mode:

1. **Host app handles authentication**: The parent application already has an authenticated IMS session
2. **Token is passed via MessagePort**: The parent sends the IMS token through `postMessage` when initializing the SDK (see `nx/utils/sdk.js:70-71`)
3. **Token is injected**: `setImsDetails(token)` stores this externally-provided token
4. **No duplicate auth**: The media library reuses the host's token instead of triggering its own IMS login

### Code flow:
```javascript
// In host app
mediaLibraryIframe.contentWindow.postMessage({ ready: true, token }, '*', [messagePort])

// In media library (nx/utils/sdk.js:70-71)
if (e.data.token) {
  setImsDetails(e.data.token);  // Inject external token
}

// Later in daFetch (nx/utils/daFetch.js:22-24)
if (imsDetails) {  // Use injected token
  const { accessToken } = await initIms();
  opts.headers.Authorization = `Bearer ${accessToken.token}`;
}
```

## nx2 Limitation

`nx2/utils/ims.js` always calls `loadIms()` which:
- Loads the Adobe IMS library directly
- Performs its own authentication flow
- Has no mechanism to accept externally-provided tokens

## Security Note

This is not a "different IMS login" - it's reusing an already-authenticated token from a trusted parent application. The host app has already completed Adobe IMS authentication; we're just accepting that token rather than forcing a second auth.

@andreituicu - Would appreciate your security review on this approach. The alternative would be to add `setImsDetails()` capability to nx2.

nx/blocks/media-library/core/paths.js
kmurugulla: ✅ **Resolved** - Reverted to using magic numbers for consistency with the rest of the codebase. Other blocks (outside media-library) also use magic numbers for HTTP status checks.

**Commit**: bcdd9b08 - refactor: use HTTP status code numbers for consistency with codebase

nx/blocks/media-library/media-library.css
kmurugulla: ✅ **Resolved** - Switched to nx2 Spectrum styles with inline `light-dark()` pattern for consistent dark mode support across all components.

Changed from custom CSS variables (`--ml-bg`, `--ml-surface`, etc.) to inline `light-dark(rgb(...), rgb(...))` values, matching the pattern used by sidebar and ensuring proper dark mode in shadow DOM components.

**Commits**:
- 830c8040 - refactor: replace custom CSS variables with inline light-dark() for consistent dark mode
- 5bc18fa4 - refactor: remove misleading CSS comment about variable inheritance

nx/blocks/media-library/ui/data.js
kmurugulla: ✅ **Resolved** - Renamed `t()` to `getMessage()` across all 13 files for better clarity since this is not i18n.

**Commit**: d1285479 - refactor: rename t() to getMessage() for clarity

nx/blocks/media-library/core/utils.js
mhaack: 🤔 For the use case of the embedding in non-DA properties I think this should either live somewhere else or in some Media Library only code, not in a shared one.

if we need special handling for IMS here for this use case this not be part of NX or DA core.

kmurugulla: in this specific use case , the media library code from da-nx is being loaded via da.live in the plugin mode. Plugin being used in layout / da editor mode. Even though plugin code is part of non-da.live related repo, as the plugin is loaded on da.live , i thought i would reuse as much as possible from da-nx. 

I am open to move the specific IMS functionalities that will enable me to reuse code between app and plugin into media library , however it will sort of creates duplication of code. And i don't know whether there are any other apps that needed this model of reuse app and plugin from same code base.
 
Plugin reference - https://github.com/adobe-rnd/aem-apps/blob/main/tools/plugins/media-library/media-library.html 

@mhaack Since I have couple of other changes added into this PR , if it is agreeable , i can take up the IMS related functionality ( moving into media library) in another PR

kmurugulla: @mhaack - [update] Internalized IMS functionality to media-library with adapter pattern commit   (https://github.com/adobe/da-nx/commit/b9129742). The adapter supports plugin mode token injection and falls back to nx2/utils/ims.js for app mode, eliminating the nx dependency.

andreituicu: _catching up with the discussion and code now..._

kmurugulla: @mhaack @andreituicu - Accenture project is planning to use the app and plugin . Since their site auth is enabled , your review on this PR will help me to make this work for that customer. Happy to address any concerns you may have .

kmurugulla:
## Plugin Migration: Moved to da-nx + Removed setImsDetails

**New Location:** `nx/public/plugins/media-library/`

**New URL:** `https://da.live/nx/public/plugins/media-library/media-library.html`

### Code Changes

**Before** (cross-origin from adobe-rnd/aem-apps):
```javascript
const { context, token } = await DA_SDK;

if (!token) {
  showError('Please register this plugin in Library and use it within authoring context.');
  throw new Error('No authentication token available');
}

// Inject token for plugin mode
const { setImsDetails } = await import('https://da.live/nx/utils/daFetch.js');
setImsDetails(token);

await import('https://da.live/nx/blocks/media-library/media-library.js');
```

**After** (same-origin in da-nx):
```javascript
const { context } = await DA_SDK;

await import('https://da.live/nx/blocks/media-library/media-library.js');
```

### Why This Works

- Same-origin iframe → shared \`localStorage\` access
- Media-library's \`initIms()\` reads directly from \`localStorage.getItem('nx-ims')\`
- Fixes 401 errors on \`.preview.da.live\` image fetches for authenticated sites

**Files:** \`nx/public/plugins/media-library/media-library.html\` (commit 3b6b65ea)

nx/public/plugins/media-library/media-library.html
```diff
@@ -0,0 +1,78 @@
...
+      // In iframe - try to load normally
+      try {
+        const { setConfig, getColorScheme } = await import('https://da.live/nx/scripts/nexter.js');
+        const DA_SDK_module = await import('https://da.live/nx/utils/sdk.js');
+        const DA_SDK = DA_SDK_module.default;
+
+        // Set color scheme for dark mode support
+        const updateColorScheme = () => {
+          document.documentElement.dataset.colorScheme = getColorScheme();
+        };
```
mhaack, line 59: `getColorScheme()` returns `light-scheme` or `dark-scheme`, but assigning that value to `data-color-scheme` has no effect because the plugin CSS has no selector for this attribute; `color-scheme: light dark` therefore continues to follow the OS setting rather than DA’s selected theme. Please apply the returned class (or set an explicit effective `color-scheme`). Also, the `#333` error text above has very low contrast on the dark `#1a1a1a` background and should use an inherited or `light-dark()` color.

nx/blocks/media-library/indexing/worker/fetch.js
```diff
@@ -674,18 +676,34 @@ export async function streamLog(
...
   let nextUrl = `${baseUrl}${separator}?${fetchParams.toString()}`;
+  let currentToken = imsToken;
 
   while (nextUrl) {
-    const resp = await workerFetchWithAuth(nextUrl, imsToken);
+    let resp = await workerFetchWithAuth(nextUrl, currentToken);
+
+    // Retry once with fresh IMS token on 401
+    if (!resp.ok && resp.status === 401) {
+      const freshImsToken = await requestTokenRefresh();
```
mhaack: This retry requests a fresh IMS token for `admin.hlx.page`, but the main-thread handler in `indexing/build.js` always responds with `getAemSiteToken().siteToken`. If IMS expires during a long build, the retry sends the AEM site token as the IMS bearer and cannot recover (often surfacing as a misleading permission error). Could we distinguish IMS-token and site-token refresh requests and refresh each credential through its corresponding provider?

nx/public/plugins/media-library/media-library.html
```diff
@@ -0,0 +1,78 @@
...
+        updateColorScheme();
+
+        // Listen for system preference changes
+        matchMedia('(prefers-color-scheme: dark)').addEventListener('change', updateColorScheme);
+
+        setConfig({ imsClientId: 'da-web', env: 'prod' });
+
+        const { context } = await DA_SDK;
```
mhaack, line 67: The SDK token initialized here never reaches the media library auth path. `nx/utils/sdk.js` stores it in `nx/utils/daFetch.js`, while media-library requests use the separate `core/ims-adapter.js` (whose `setImsDetails()` currently has no caller), and worker builds read only `window.adobeIMS`. When the SDK token is the iframe’s only credential, API calls are sent without `Authorization`, indexing can fail with `No IMS token available`, and SDK token refreshes update the wrong module. Existing `nx-ims` state may mask this in some sessions, making the failure environment-dependent. Please inject SDK token updates into the media-library adapter and expose one shared token getter for worker startup.

nx/blocks/media-library/indexing/worker/fetch.js
mhaack: @andreituicu any thoughts on this?

kmurugulla: The worker now distinguishes between token types:
  - `requestTokenRefresh('ims')` for `admin.hlx.page` APIs (audit log, medialog, status)
  - `requestTokenRefresh('site')` for `.aem.page` content fetches

  **Changes:**
  - Worker: `indexing/worker/fetch.js` passes `tokenType` parameter
  - Main thread: `indexing/build.js` routes to correct provider:
    - `tokenType === 'ims'` → calls `initIms()` to refresh IMS token
    - `tokenType === 'site'` → calls `getAemSiteToken()` to refresh site token

  This ensures admin.hlx.page 401 retries get a fresh IMS token instead of incorrectly receiving the site
  token.

  ---
  Commit: acc7da5b - fix: distinguish IMS and site token refresh in worker to prevent auth failures on
  admin.hlx.page

nx/public/plugins/media-library/media-library.html
kmurugulla: This should be fixed now

nx/public/plugins/media-library/media-library.html
kmurugulla: Looking for clarification @mhaack 

**Context:**
   - I removed the `setImsDetails(token)` call because I thought same-origin would allow media-library to
  access parent's `localStorage.getItem('nx-ims')` directly

  **Your comment says:**
  > "Please inject SDK token updates into the media-library adapter"

  **Questions:**

  1. **Should we restore the token injection?** Even with same-origin, does the SDK token need to be
  explicitly injected into `core/ims-adapter.js`?

  2. **What's the correct import path?** Should it be:
     ```javascript
     // Option A: Media-library's own adapter
     const { setImsDetails } = await import('https://da.live/nx/blocks/media-library/core/ims-adapter.js');

     // Option B: Shared daFetch module (original path)
     const { setImsDetails } = await import('https://da.live/nx/utils/daFetch.js');

  3. Why doesn't same-origin eliminate the need? I thought moving to da-nx would let initIms() read from
  shared localStorage. What am I missing about the SDK token vs localStorage IMS session?
  4. "Expose one shared token getter for worker startup" - Does this mean we need a new export from
  core/ims-adapter.js that the worker can call to get the current token?

mhaack: Actually "I removed the setImsDetails(token)" is correct.

> Why doesn't same-origin eliminate the need? I thought moving to da-nx would let initIms() read from
shared localStorage. What am I missing about the SDK token vs localStorage IMS session?

You do not need to read the IMS directly from local storage. `const { context } = await DA_SDK;` is the correct way to do it here as you only use org/site in the `media-library.html` no IMS token, I ignore my comment here.

mhaack (approved):

PR #583, opened 2026-07-08
Locnx2

### LINKED ISSUE #532: [nx2] loc

ravuthu:

Fix #532 

Test URLs:


.husky/pre-commit
```diff
@@ -1,2 +1 @@
 npx lint-staged
-npm test
```
hannessolo, line 2: Seems unintentional?

nx/blocks/loc/project/views/sync.js
```diff
@@ -1,19 +1,18 @@
...
-const style = await getStyle(import.meta.url);
-const shared = await getStyle(`${nxBase}/blocks/loc/project/views/shared.js`);
-const buttons = await getStyle(`${nxBase}/styles/buttons.js`);
+const { nxBase: nx } = config;
+const style = await loadStyle(import.meta.url);
 
 const ICONS = [
-  `${nxBase}/blocks/loc/img/Smock_Checkmark_18_N.svg`,
-  `${nxBase}/blocks/loc/img/Smock_ChevronRight_18_N.svg`,
+  `${nx}/public/icons/Smock_Checkmark_18_N.svg`,
```
hannessolo: Optional: could update this to the new way of loading icons (<svg><use> directly from da.live)

nx/blocks/loc/project/views/translate.js
```diff
@@ -1,19 +1,17 @@
...
-const { nxBase } = getConfig();
-const style = await getStyle(import.meta.url);
-const shared = await getStyle(`${nxBase}/blocks/loc/project/views/shared.css`);
-const buttons = await getStyle(`${nxBase}/styles/buttons.js`);
+const { nxBase: nx } = config;
+const style = await loadStyle(import.meta.url);
 
 const ICONS = [
-  `${nxBase}/blocks/loc/img/Smock_ChevronRight_18_N.svg`,
+  `${nx}/public/icons/Smock_ChevronRight_18_N.svg`,
```
hannessolo: Same here

nx/blocks/loc/project/index.js
```diff
@@ -163,6 +163,43 @@ const getDaUrl = (url) => {
   return { org, repo, pathname };
 };
 
+function replaceHtml(text, fromOrg, fromRepo, options = {}) {
```
hannessolo, line 166: Can you double-check whether we already have a similar function in nx2, and if yes move to a shared utils location?

nx/blocks/loc/views/dashboard/buttons.css
```diff
@@ -0,0 +1,11 @@
+button,
```
hannessolo, line 1: Could we potentially use the styles from nx2/styles/form.css instead?

nx/blocks/loc/views/options/options.js
```diff
@@ -1,18 +1,19 @@
...
-const { nxBase: nx } = getConfig();
+const { nxBase: nx } = nxConfig;
 
-const style = await getStyle(import.meta.url);
+const style = await loadStyle(import.meta.url);
 
 const ICONS = [
-  `${nx}/blocks/loc/img/Smock_Close_18_N.svg`,
-  `${nx}/blocks/loc/img/Smock_Add_18_N.svg`,
+  `${nx}/public/icons/Smock_Close_18_N.svg`,
```
hannessolo: See comments about icons

nx2/scripts/nx.js
```diff
@@ -102,6 +102,31 @@ export const loc = ([first], ...values) => {
   return strings.get(key) ?? key;
 };
 
+export const loadStyle = (() => {
```
hannessolo: Is this move intentional?

nx2/utils/nxToggle.js
```diff
@@ -0,0 +1,14 @@
+const config = await (async () => {
```
hannessolo: What does this do? Is it just for testing or do we need it? The goal is for evertyhing to be nx2 only.

hannessolo (commented):
Since you added icons to the code in this PR, I think it would be a good opportunity to switch the icon loading methods - load them directly from da.live/img/icons/... instead.

.husky/pre-commit
ravuthu: Looks like it was unintentional. Added it back. @auniverseaway, hope you are good with that.

nx/blocks/loc/project/views/sync.js
ravuthu: updated

nx/blocks/loc/project/views/translate.js
ravuthu: updated

nx/blocks/loc/project/index.js
ravuthu: I didn't find it in nx2. I believe that the `project/` folder was used for legacy Rollout and Sync. So, it didn't make sense to move it to common utils. Let me know if you still want to move it.

nx/blocks/loc/views/dashboard/buttons.css
ravuthu: I don't think it makes sense to import all of the form styles just to reset the buttons. What do you think?

nx/blocks/loc/views/options/options.js
ravuthu: updated

nx2/scripts/nx.js
ravuthu: @auniverseaway, not sure if this was intentional on your part.

nx2/utils/nxToggle.js
ravuthu: @auniverseaway , not sure if we need this at all. I updated the utils.js (https://github.com/adobe/da-nx/pull/583/changes/f980af14abc448b0b497084f5dcf9182d5638667#diff-e3c0e4a8446171228c63c5f2441f0fdb889f13af246617ff25d6c3e4cd8bbd85R216) to load from a relative path instead of using nx variable. The only other place this was used is `nx/blocks/loc/project/views/translate.js` which is the legacy implementation.

janaki-r-bhagwath (dismissed):

nx2/utils/nxToggle.js
hannessolo: I think this is unused if I saw correctly - can you double check and if yes, remove the file

nx2/scripts/nx.js
hannessolo: Doesn't really matter from a code POV, but could you move it back so we keep the history of the file cleaner.

ravuthu: Updated: one thing to note is that the `cache` is moved from module-level into the `loadStyle` IIFE, so it's now private to `loadStyle`

nx2/utils/nxToggle.js
ravuthu: updated

nx2/scripts/nx.js
```diff
@@ -335,27 +335,28 @@ export async function loadArea({ area } = { area: document }) {
   }
 }
 
-const cache = {};
+export const loadStyle = (() => {
```
hannessolo: do we really need this in an IIFE? I'd prefer to have no changes to this code if possible

hannessolo (commented):
As I understand this has been tested by teams who use the feature?

Code wise it looks good now except my comment in /scripts/nx.js

nx2/scripts/nx.js
ravuthu: Not mandatory, but it would be in sync with other implementations in the same file. Reverted it for now

ravuthu:
> As I understand this has been tested by teams who use the feature?

I am the only one verified it. Not sure who else will be testing this, or if there is a process that I am not aware of.

hannessolo (approved):

PR #585, opened 2026-07-08
chore: update styling for multi-line nx-menu items with description

usman-khalid:
Small styling adjustments for EW's slash menu to support multi-line items needed for https://github.com/adobe/da-live/pull/1084

### DIFF
--- a/nx2/blocks/shared/menu/menu.css
+++ b/nx2/blocks/shared/menu/menu.css
@@ -41,6 +41,15 @@
   background: var(--s2-gray-200);
 }
 
+.menu-item:disabled {
+  color: light-dark(var(--s2-gray-500), var(--s2-gray-600));
+  cursor: default;
+}
+
+.menu-item:disabled:hover {
+  background: none;
+}
+
 .menu-item-icon {
   display: flex;
   align-items: center;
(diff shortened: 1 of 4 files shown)


nx2/blocks/shared/menu/menu.css
```diff
@@ -29,7 +29,7 @@
   background: none;
   color: var(--s2-gray-900);
   font-family: inherit;
-  font-size: var(--nx-menu-font-size, var(--s2-body-size-xs));
+  font-size: var(--s2-body-size-xs);
```
sharanyavinod: Why remove this? This allows consumers to customize menu font size if needed and is already used in da-live

usman-khalid: Ah, thanks. It was undefined so I thought it was a broken var. Brought it back.

sharanyavinod (commented):
Could you please update [the doc](https://github.com/adobe/da-nx/blob/main/docs/menu.md) to reflect the latest state?

nx2/blocks/shared/menu/menu.js
```diff
@@ -119,9 +120,11 @@ class NxMenu extends LitElement {
           data-id=${item.id}
           class="menu-item ${item.id === this._active ? 'menu-item-active' : ''}"
           type="button"
+          ?disabled=${item.disabled}
+          aria-disabled=${item.disabled ? 'true' : nothing}
```
sharanyavinod, line 124: You can skip the aria-disabled here. For buttons, the native disabled attibute is sufficient.

usman-khalid:
Not needed anymore until we explicitly need support for disabled buttons in nx-menu for something else.

PR #586, opened 2026-07-09
feat(nx2): port sandbox org-check to nx2

### LINKED ISSUE #539: [nx2] toast
Is this still needed?

hannessolo:
* Adds back the sandbox org warning
* Uses toast from nx2

Fix #539 

https://da.live/#/da-sites/nexter
- After: https://main--da-live--adobe.aem.live/?nx=feat-nx2-org-check#/da-sites/nexter

### DIFF
--- a/nx2/blocks/shared/toast/toast.js
+++ b/nx2/blocks/shared/toast/toast.js
@@ -4,6 +4,7 @@ import { loadHrefSvg } from '../../../utils/svg.js';
 
 export const VARIANT_SUCCESS = 'success';
 export const VARIANT_ERROR = 'error';
+export const VARIANT_WARNING = 'warning';
 
 const CLOSE_ICON_URL = new URL('../../../img/icons/S2_Icon_Close_20_N.svg', import.meta.url).href;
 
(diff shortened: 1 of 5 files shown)


nx2/blocks/shared/toast/toast.js
```diff
@@ -75,12 +79,15 @@ class NxToast extends LitElement {
     const text = this.message?.trim();
     if (!text) return nothing;
     const isError = this.variant === VARIANT_ERROR;
+    const isWarning = this.variant === VARIANT_WARNING;
+    let variantClass = VARIANT_SUCCESS;
+    if (isError) variantClass = VARIANT_ERROR;
+    else if (isWarning) variantClass = VARIANT_WARNING;
```
mhaack: Only cosmetic but this could be ...

```suggestion
    const variantClass = isError ? VARIANT_ERROR : isWarning ? VARIANT_WARNING : VARIANT_SUCCESS;
```

mhaack (commented):
Does the `toast.js` has some tests as well we should update or should we add some.

mhaack (approved):

PR #587, opened 2026-07-09
Feat/tree block nx2 migration

### LINKED ISSUE #540: [nx2] tree (traverse)

hannessolo:
fix #540 

https://main--da-live--adobe.aem.live/apps/traverse?nx=feat-tree-block-nx2-migration&nxver=2

### DIFF
--- a/nx/blocks/tree/tree.css
+++ b/nx/blocks/tree/tree.css
@@ -1,20 +1,20 @@
 :host {
   display: block;
-  max-width: var(--grid-container-width);
-  margin: var(--spacing-500) auto;
+  max-width: var(--se-grid-container-width, var(--grid-container-width));
+  margin: var(--s2-spacing-500, var(--spacing-500)) auto;
 }
 
 form {
   display: flex;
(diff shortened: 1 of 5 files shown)


nx/blocks/tree/tree.css
```diff
@@ -1,20 +1,20 @@
...
 }
 
 form {
   display: flex;
-  gap: var(--spacing-400);
+  gap: var(--s2-spacing-400, var(--spacing-400));
 }
 
 input {
-  font-family: var(--body-font-family);
```
mhaack, line 13: For the form fields we can/should use https://github.com/adobe/da-nx/blob/main/nx2/styles/form.css and can reduce this here.

nx/styles/buttons.css
mhaack, line 1: Same for this button styles, they can be replaced by https://github.com/adobe/da-nx/blob/main/nx2/styles/form.css

mhaack (dismissed):

sharanyavinod (approved):

PR #591, opened 2026-07-09
Add ASO constants workflow and keywords (updated) columns.

janaki-r-bhagwath:
- Keep fixed copy in per-page -constants file with {{slug}} tokens in listings so translation leaves legal/fixed text unchanged
- Resolve constants at preview, copy, and export
- Extend keywords JSON generation with companion "(updated)" columns for GLaaS handoff

### DIFF
--- a/nx/blocks/loc/connectors/glaas/translationMetadata.js
+++ b/nx/blocks/loc/connectors/glaas/translationMetadata.js
@@ -107,32 +107,84 @@ export function needsKeywordsMetadata(parsedSchema) {
   return Object.values(parsedSchema).some(hasKeywords);
 }
 
-export async function fetchKeywordsFile(org, site, pagePath) {
-  // Remove .html extension if present and add -keywords.json
+const CONSTANT_SLUG_PATTERN = /^[a-z0-9]+(-[a-z0-9]+)*$/;
+const CONSTANT_TOKEN_PATTERN = /\{\{([a-z0-9]+(-[a-z0-9]+)*)\}\}/g;
+
+function metadataPathForPage(pagePath, suffix) {
   const cleanPath = pagePath.replace(/\.html$/, '');
-  const keywordsPath = `${cleanPath}-keywords.json`;
-  // Try primary path
-  let url = `${DA_ORIGIN}/source/${org}/${site}${keywordsPath}`;
+  return `${cleanPath}${suffix}`;
+}
+
+async function fetchMetadataFile(org, site, pagePath, suffix, readBody) {
+  const metadataPath = metadataPathForPage(pagePath, suffix);
+  let url = `${DA_ORIGIN}/source/${org}/${site}${metadataPath}`;
   try {
-    const resp = await daFetch(url);
+    let resp = await daFetch(url);
     if (resp.ok) {
-      return resp.json();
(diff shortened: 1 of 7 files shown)


janaki-r-bhagwath:
@chrischrischris @jenssingler , Raising the PR so it can be reviewed and approved. Kindly do not merge - since it there is a non backward compatible change on the GLaaS side and I would need to merge this PR in sync with them.

nx/blocks/loc/connectors/glaas/translationMetadata.js
```diff
@@ -107,34 +107,86 @@ export function needsKeywordsMetadata(parsedSchema) {
...
+    const slug = match[1];
+    if (CONSTANT_SLUG_PATTERN.test(slug)) slugs.add(slug);
+    match = pattern.exec(text);
+  }
+  return [...slugs].sort();
+}
+
+function slugFromConstantsBlock(block) {
+  if (!block) return null;
+  const slugClasses = [...block.classList].filter((className) => className !== 'aso-constants');
```
jenssingler: This is ASO specific and shouldn't be in da-loc.

nx/blocks/loc/connectors/glaas/translationMetadata.js
```diff
@@ -107,34 +107,86 @@ export function needsKeywordsMetadata(parsedSchema) {
...
+  if (slugClasses.length !== 1) return null;
+  const slug = slugClasses[0];
+  return CONSTANT_SLUG_PATTERN.test(slug) ? slug : null;
+}
+
+function constantsRowsFromHtml(html) {
+  if (!html || typeof html !== 'string') return [];
+  const doc = new DOMParser().parseFromString(html, 'text/html');
+  const byLanguage = new Map();
+  doc.querySelectorAll('div.aso-constants').forEach((block) => {
```
jenssingler: This is ASO specific and shouldn't be in da-loc.

jenssingler (changes requested):
I don't think that having ASO specific class names in the da-loc code is desirable.
Wouldn't a similar mechanism like keywords from metadata or config work for here as well? 
In addition (even though you added comments) this excessive console logging is not desirable in PROD.
And your description says "Add a Sidekick constants authoring tool with DOM substitution that preserves section breaks" - which I don't see here.

janaki-r-bhagwath:
@jenssingler , address the aso-constants and the removed logs.

jenssingler (changes requested):
> address the aso-constants and the removed logs.

I still see all (most) of the logs in there...

Also in `addTranslationMetadata` I think there are three places where the same document is fed to the parser and then processed: `annotateHTML`, `fieldConstantSlugs` and `buildLanguageMetadata`. With enough urls doing this just once could make a noticeable difference.
`annotateHTML` would return the doc instead of the `body.innerHTML` string and this can be passed to `fieldConstantSlugs` and `buildLanguageMetadata`.
`annotateHTML` and  `buildLanguageMetadata` are exported, but AFAICS only for unit test access. Those tests would have to be adjusted, but no dependent code changes that I can see.

jenssingler (approved):
Looks good to me now. Thanks for addressing all my comments...

PR #592, opened 2026-07-10
feat(ew): show loading spinner on preview & publish

usman-khalid:
Currently there's no visual cue that something is in progress when a page is previewed or published from the paper airplane in `ew-actions`.

Adds a loading spinner.

### DIFF
--- a/nx2/blocks/ew-actions/ew-actions.css
+++ b/nx2/blocks/ew-actions/ew-actions.css
@@ -106,6 +106,32 @@
   height: 18px;
 }
 
+.preview-dropdown-btn.is-busy:disabled {
+  color: light-dark(var(--s2-gray-25), #fff);
+  background-color: var(--s2-blue-900);
+}
+
+.preview-dropdown-spinner {
+  display: block;
+  flex-shrink: 0;
+  width: 12px;
+  height: 12px;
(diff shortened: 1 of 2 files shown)


mhaack:
I like the idea, but instead of the spinner, should we use the flying paper plane like we have in the stock base editor? @sharanyavinod any thoughts?

usman-khalid:
> I like the idea, but instead of the spinner, should we use the flying paper plane like we have in the stock base editor? @sharanyavinod any thoughts?

I know that's what we currently have in DA, but the spinner is standard for loading experiences in Spectrum: https://opensource.adobe.com/spectrum-web-components/components/progress-circle/

mhaack:
ACK, lets keep the spinner then

mhaack (approved):

sharanyavinod (approved):

sharanyavinod:
We have multiple spinners across components now, so this would anyway need to be extracted to be a common, reusable class. For now, can stay as is. Thanks for the contribution!

usman-khalid:
> We have multiple spinners across components now, so this would anyway need to be extracted to be a common, reusable class. For now, can stay as is. Thanks for the contribution!

Created https://github.com/adobe/da-nx/issues/599 to track

PR #596, opened 2026-07-13
feat(chat): include browser timezone in page context for agent

amol-anand:
inject user's timezone to the agent context so the agent can use that for time-based tasks and schedule publish MCP tool as part of aem-agentic-plugins (that also has a get current time tool associated with it)

This PR needs to be merged in conjunction with the one in da-agent: https://github.com/adobe-rnd/da-agent/pull/58

### DIFF
--- a/nx2/blocks/chat/chat-controller.js
+++ b/nx2/blocks/chat/chat-controller.js
@@ -90,7 +90,15 @@ export default class ChatController {
 
   _pageContextForAgent() {
     const { org, site, path, view } = this._context ?? {};
-    return org && site ? { org, site, path: path ?? '', view } : undefined;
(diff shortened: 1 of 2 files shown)


mhaack:
Do we have a use case for this? Where or in which situation would this be used?

amol-anand:
when using schedule publish, when the user says publish this page in 4 hours or tomorrow at 10 am, we currently have a tool to get the agent the current time, but then the agent then needs to ask the user for the timezone so it scheduled it accurately against the user's local time. So injecting the current user's timezone to the agent context would give it the information it needed without having to ask the user their current timezone.

mhaack (approved):

PR #605, opened 2026-07-15
feat(ew): add layout-mode block & image selection

usman-khalid:
Should be merged with https://github.com/adobe/da-live/pull/1114

Adds support for block and image selection in EW canvas.


nx/public/plugins/quick-edit/quick-edit.css
```diff
@@ -271,3 +275,65 @@ da-palette {
...
+  pointer-events: auto;
+  cursor: pointer;
+}
+
+.qe-pill-grip {
+  width: 10px;
+  height: 12px;
+  flex-shrink: 0;
+  background-color: currentcolor;
+  mask: url("https://da.live/img/icons/s2-icon-select-20-n.svg") no-repeat center / contain;
```
sharanyavinod: Since this runs on the customer side, might be better to use via relative path on nx similar to how close is used in the same file. Wdyt?

sharanyavinod (commented):
Minor question, else lgtm.

nx/public/plugins/quick-edit/quick-edit.css
usman-khalid: Done.

sharanyavinod (approved):

Issue #599, opened 2026-07-14
Extract a common, reusable loading spinner (nx2)


usman-khalid:
Follow-up from review feedback on #592 

nx2 has multiple near-identical rotating border-spinners, each defining its own `@keyframes`:

- `nx2/blocks/ew-actions/ew-actions.css` — `.preview-dropdown-spinner` + `@keyframes preview-dropdown-spin` (added in #592)
- `nx2/styles/form.css` — `.da-checkbox.loading` + `@keyframes da-spin`
- `nx2/public/plugins/quick-edit/quick-edit.css` — `picture.image-uploading::after` + `@keyframes spin`

All three are the same pattern (border + `border-top-color` transparent + `border-radius: 50%` + `rotate(360deg)`), just with different sizes/colors and duplicated keyframes.

Should consolidate and re-use across these UI surfaces.

PR #610, opened 2026-07-20
doc: introduce style guide

sharanyavinod:
https://stlegde--da-nx--adobe.aem.page/docs/style-guide/style-guide.html

Also includes fix for #599


docs/style-guide/style-guide.html
hannessolo, line 1: should we move all this under docs too? 

That would also match the pattern that's commonly used for the block library - which this is similar to - which in the content is under /docs/library/blocks.

nx2/styles/buttons.css
hannessolo, line 1: Can't really tell from the diff - does the form.css file still have styles for buttons, or will we always need to include this one too?

Also, should we be doing refactoring of the code to a) switch places we manually styled buttons to import this instead and b) siwtch places we used form.css to use this instead?

hannessolo (commented):
Looks really nice!

I also wonder if we should add somethign to docs/agents.md that says "when you create a new component/modify it, update the style guide"?

nx2/styles/buttons.css
sharanyavinod: > does the form.css file still have styles for buttons

I moved all the buttons to buttons.css just to keep it consistent - good or too confusing?

> switch places we manually styled buttons to import this instead and b) siwtch places we used form.css to use this instead

Definitely, would be follow up when this is good and merged

sharanyavinod:
> docs/agents.md that says "when you create a new component/modify it, update the style guide"

Makes sense, I'll add that as well. And should we update the contributing md to include this? I was a bit hesitant since this is quite ew specific atm, but maybe its the way to go. Thoughts?

docs/style-guide/style-guide.html
sharanyavinod: I dont have strong opinions here, we can move it already or later once its a bit fleshed out. Whichever you think makes more sense.

sharanyavinod: On second thought, maybe its good to keep it here so its easier to keep up to date and in sync whenever changes are made in any of the related files. Wdyt?

hannessolo: > its good to keep it here

Keep it here being where? I would definitely keep it in this repo. I just thought to move it under /docs in this codebase instead of /public - or you prefer not to do that?

sharanyavinod: I like the idea of the link being reachable and accessible as is, hence why leaning a bit more towards having it in public.Since our goal is to have as much people using this, and discovering this, having an accessible link might make it easier and nicer to link in any contribution docs we share etc. Unless I miss something and its not as straightforward with this approach either, or feels overkill?

hannessolo: By reachable and accessible you mean that the link https://stlegde--da-nx--adobe.aem.page/nx2/public/style-guide/style-guide.html works? That's not exclusive to the /public directory, static html can be served from anywhere.

I guess by convention most projects would put html in public, but in this case public is more intended to mean "public to import - stable apis that won't change" than publicly accessible, since anything committed to git in EDS projects is anyway publicly accessible.

sharanyavinod: > but in this case public is more intended to mean "public to import - stable apis that won't change" than publicly accessible

Fair enough, I'll move it to under docs

hannessolo (approved):

PR #614, opened 2026-07-23
refactor: unify panel events

sharanyavinod:
Changes introduced: 
- All panel state exposed via events, for consistency: PANEL_EVENT.OPEN / PANEL_EVENT.CLOSE (nx-panel-open / nx-panel-close) are now the single, canonical way to open or close any panel from anywhere — headers, toolbars, BYO plugin iframes — without importing panel.js directly. This replaces the mix of ad hoc events and direct function calls each panel type used before.
- Panels are addressed by section, not position. PANEL_EVENT.OPEN's payload is { section, id?, options? }: section is the stable name a host page registers ('chat', 'tools'); id selects a specific view within a multi-view section (e.g. a tool-panel tab); options is opaque, passed straight through to that section's own onShow handler to interpret however it needs (e.g. chat's pre-fill text + autoSend).
- Position is now just a registration-time detail, not the identity.
- Panel open/closed state is now unified across pages. Previously canvas persisted its panel state in localStorage (surviving restarts) while browse and form each tracked their own separate sessionStorage flag (cleared per tab). All three now read the same wasPanelOpen(section)-backed store — opening chat in canvas keeps it open when you navigate to browse or form, and vice versa.
- Removed the old fragment-based panel mechanism (action-button.js, restorePanels(), loadPanelContent()) — it predated the section registry, had no confirmed usage anywhere in either repo's content, and left openPanel's fragment branch as dead weight once removed.
- All event names shared betweed da-nx and da-live for chat and panel are now shared constants

Relates to https://github.com/adobe/da-live/pull/1170

https://ewpnl--da-live--adobe.aem.live/canvas?nx=ewpnl#/exp-workspace/frescopa/coffee


hannessolo (approved):
Very nice

PR #615, opened 2026-07-23
feat: Merge DA-legacy and hlx6 source-bus sites in org-level listing

### LINKED ISSUE #1169: [HLX6] Mix source bus listing with DA listing (org level)
The org-level site list (`/{org}`, no site) should include sites from source bus (hlx6) in addition to sites from DA content.

**Problem**

Org-level listing currently only surfaces DA-legacy folders via `${DA_ADMIN}/list/{org}` (see `getDaApiPath`'s `LIST` branch in `nx2/utils/api.js`). Source-bus (hlx6) sites are invisible at the org level — there's no equivalent lookup wired in.

**Source bus endpoint**

Use `https://api.aem.live/{org}/source/` to read the source-bus site list. Note: there's an existing unused stub `org.listSites({ org })` in `nx2/utils/api.js` that calls `${AEM_API}/${org}/sites` — that path is wrong and should be corrected to `${AEM_API}/{org}/source/`.

**Where the fix lives**

Primarily in the `da-nx` repo (`nx2/utils/api.js`: `getDaApiPath`'s `LIST` branch, and the unused `org.listSites`). da-live's `test/fixtures/nx2/utils/api.js` mirrors this file for unit tests and needs to be kept in sync with whatever changes land in da-nx.

**Expected behavior**

Combine/dedupe the DA-legacy and source-bus site lists into a single merged result for org-level consumption (`da-sites.js` / `da-list.js`).

tripodsan:
## Summary
- `source.list({ org })` (no `site`) only surfaced DA-legacy folders via `${DA_ADMIN}/list/{org}`, so hlx6-upgraded sites were invisible in org-level listings (e.g. `/{org}` in da-live). It now queries both the DA-legacy list and the hlx6 source-bus site list in parallel, normalizes both to the shared item shape, and dedupes by name (legacy entry wins on collision). `ok` is true if either side succeeds, so a 404 on one side (non-migrated org, or a fully-migrated org with no legacy content) doesn't blank the other's results.
- Fixed `org.listSites`, which was hitting the wrong/unused stub endpoint `${AEM_API}/{org}/sites`; corrected to the documented source-bus endpoint `${AEM_API}/{org}/source/`.
- Since only DA paginates (hlx6 has no continuation token), `org.listSites` is now only queried on the first page of an org-level list, instead of being redundantly re-fetched and re-merged on every page.
- Updated `api.d.ts` / `api.md` to document the merge behavior.

Fixes adobe/da-live#1169.

Note: `da-live`'s `test/fixtures/nx2/utils/api.js` mirrors this file for its own unit tests and will need a manual sync in that repo — out of scope here since it lives in a separate repo.

## Test plan
- [x] `npm test` — full suite passes (1097/1097)
- [x] `npm run lint:js` — clean on touched files
- [x] Added unit tests for: merge + dedupe, DA-legacy 404 (hlx6-only fallback), hlx6 404 (legacy-only fallback), both failing (`ok: false`), and hlx6 site list only fetched on the first page (skipped when `continuationToken` is present)
- [x] Fixed existing tests asserting the old (incorrect) `org.listSites` URL and single-fetch org-only listing behavior


nx2/utils/api.js
```diff
@@ -206,21 +206,43 @@ export const source = {
...
     const cleanPath = (path || '').replace(/\/$/, '');
     const parentPath = `/${org}${site ? `/${site}` : ''}${cleanPath}`;
     const fetchOpts = continuationToken
       ? { ...opts, headers: { ...opts?.headers, 'da-continuation-token': continuationToken } }
       : opts;
+
+    if (!site) {
+      // Only DA returns a continuation token; hlx6 has no pagination, so its
+      // (unpaginated) site list is only fetched on the first page.
+      const [legacyResp, sitesResp] = await Promise.all([
```
benpeter: IIUC this would fail both if either fails - is that intended or should this be `Promise.allSettled` ?

tripodsan: good point. thanks.

benpeter (approved):
lgtm

hannessolo (approved):

PR #621, opened 2026-07-27
fix: reuse styles in header

sharanyavinod:
Updated button styles and adapted header to reuse common styles.

https://ewhead--da-nx--adobe.aem.live/docs/style-guide/style-guide.html#form
https://da.live/canvas?nx=ewhead&nxver=2#/exp-workspace/frescopa/coffee

Requires https://github.com/adobe/da-live/pull/1182 in parallel - merged


hannessolo:
Screenshot 2026-07-31 at 11 26 19

Here some buttons look broken

hannessolo:
Screenshot 2026-07-31 at 11 30 14

 I think having different font sizes here looks strange - I'd rather have everything in 14px like the breadcrumb. WDYT?

sharanyavinod:
I was thinking of tackling breadcrumb separately, since that has some alignment issues that need a closer look and the chevron which needs to be adjusted. 
Regarding the sizing, I agree they do look at odds with each other now - but changing the new ones would mean all buttons throughout the page would need to be updated in order to be consistent. So rather, it might be the breadcrumb itself that needs modification. Let me talk to Kilian - since just dropping to 12px in the breadcrumb seems too small to me.

sharanyavinod:
@hannessolo Discussed with Kilian and updated defaults to 14px

hannessolo (approved):

PR #626, opened 2026-07-29
feat: use api.aem.live to store site config

tripodsan:
Stores the DA config directly in the helix6 config below the 'editor' object. it simplifies the sheetbased config by removing the sheet meta information. eg:

```
  "editor": {
    "da": {
      "library": [
        {
          "name": "My Library",
          "path": "/tools/lib1.html"
        },
        {
          "name": "My Other Lib",
          "path": "/tools/lib2.html"
        }
      ],
      "apps": [
        {
        ....
        }
      ]
    }
    ```

### CODE REVIEW

File: nx2/utils/api.js [hunk 1]
```diff
@@ -74,6 +76,16 @@ export const aem = {
 // config: top-level org/site config.
 export const config = {
   get: withArgs(async ({ org, site, cachebust }) => {
+    const hlx6 = await isHlx6(org, site);
+    if (hlx6) {
+      const url = `${AEM_API}/${org}/sites/${site}/config/editor/da.json`;
+      const resp = await daFetch({ url });
```
anfibiacreativa, line 82: `get`/`save`/`delete` all repeat the same `await isHlx6(org, site)` gate and the same `${AEM_API}/${org}/sites/${site}/config/editor/da.json` path. maybe we pull just the url into a small `hlx6ConfigUrl(org, site)` helper? i’d keep the request handling separate, since the branches are intentionally different.

nx2/utils/api.js
```diff
@@ -74,6 +76,16 @@ export const aem = {
...
+      if (resp.ok) {
+        const cfg = object2sheet(await resp.json());
+        resp.json = () => cfg;
```
anfibiacreativa, line 85: we're mutating a real Response and making json() return a non-promise. it works because callers await it, but it's fragile. why not return a small shim { ok: resp.ok, status: resp.status, json: async () => cfg } instead of patching the live Response?

tripodsan: I just copied the pattern from `withSourceJson` below.

nx2/utils/api.js
tripodsan: I don't like those helpers too much. if you look at the code, you never see what's really going on. this is more verbose but also more clear of what happens. in the end you have a helper for every little string that you need to compose.

nx2/utils/utils.js
```diff
@@ -153,3 +153,37 @@ export const loadPageStyle = (href) => new Promise((resolve) => {
 
 export { loadStyle };
 export { default as loadScript } from '../../nx/utils/script.js';
+
+// Sheet-format JSON (as served by DA/AEM config stores) -> simple object,
+// keyed by sheet name(s), values are each sheet's `data` array.
+export function sheet2object(json) {
```
kptdobe, line 159: I assume this should go into `da-parser` (at one point...).

kptdobe (approved):

nx2/utils/utils.js
tripodsan: maybe... but I doubt that we need it elsewhere.

PR #634, opened 2026-08-03
fix(nx-loc-url-details): calculate and use correct editor path in preflight edit links

shsteimer:
WIP: since this makes `getEditPath` async, and there is also a very similar function in https://github.com/adobe/da-live/blob/main/blocks/browse/shared.js, sorta feels like we should make this a single global util, and/or pre-load the config so that function need not be async.

...To be Continued...

Fix #<gh-issue-id>

Test URLs:

### DIFF
--- a/nx/blocks/loc/views/url-details/index.js
+++ b/nx/blocks/loc/views/url-details/index.js
@@ -21,7 +21,7 @@ function splitPath(path) {
 // Mirrors da-live's da-browse.js getEditor(): EW flag picks the canvas/edit
 // default, then the site's `editor.path` config rows override by longest
 // path-prefix match, so linked-page edit URLs land in the same editor browse would use.
-async function getEditorRoute(org, site, path) {
+async function getEditorRoute({ org, site, path }) {
   const isEW = await isEWEnabled({ org, site });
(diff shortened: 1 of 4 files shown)


nx/blocks/loc/views/url-details/index.js
```diff
@@ -16,12 +18,43 @@ function splitPath(path) {
...
+  const defRoute = isEW ? '/canvas#' : '/edit#';
+
+  const configs = await Promise.all(fetchDaConfigs({ org, site }));
+  const rows = configs.filter(Boolean).reverse().flatMap((c) => getFirstSheet(c) || []);
+  const editorConfs = rows.reduce((acc, row) => {
+    if (row.key === 'editor.path') acc.push(row.value);
+    return acc;
+  }, []);
+
+  const matchedConfs = editorConfs.filter((conf) => path.startsWith(conf.split('=')[0]));
```
anfibiacreativa, line 35: split('=')[1] drops everything after a second =, so an editor.path route with a query string breaks. please parse with indexOf/slice

anfibiacreativa (changes requested):

PR #642, opened 2026-08-05
feat(schema-editor): validate schemas and surface errors clearly

kozmaadrian:
## Summary
Overhauls the schema editor's validation and error experience, and adopts `@adobe/da-sc-sdk@0.3.0` (whose `schemaIssues` now carry a machine `reason`, a human `message`, a schema-source `schemaPath`, a data `pointer`, and a structured `details` blob).

## What changed

**Schema editor**
- **Resilient loading** — a schema saved with invalid JSON no longer breaks the
  editor. The broken schema opens with its raw text for repair instead of taking
  down the whole schema list.
- **Validation on save and on select** — invalid JSON is rejected, and structural
  errors the form engine can't compile are blocked before saving. The default
  template is seeded valid so a brand-new schema isn't blocked.
- **Dedicated errors panel** above the editor — one issue per line, each shown with
  its human message and the schema-source location (e.g.
  `The type must be a single value, not an array (at #/$defs/YearAchievement)`).
- **Inline JSON linting** — JSON syntax errors are underlined on the offending line
  via CodeMirror's linter.

**Dependencies**
- `@adobe/da-sc-sdk` bumped to `^0.3.0`; the vendored bundle is rebuilt from the
  published release and re-exports `validateSchema`.
- The vendored CodeMirror bundle now exposes the JSON linter and lint gutter.

**Form block**
- `describeIssue` reads issue context from the SDK's `details` (composition keyword /
  type), matching the restructured `schemaIssues` shape.

Test URLs:
- https://da.live/apps/schema?nx=schmfix#/exp-workspace/frescopa > Select existing schema `Unsupported Schema Test`

Screenshot 2026-08-05 at 3 07 45 pm


nx/blocks/schema-editor/utils/utils.js
```diff
@@ -29,8 +32,15 @@ async function loadSchema(schema) {
...
-  const jsonStr = dom.querySelector('code').textContent;
-  return JSON.parse(jsonStr);
+  const jsonStr = dom.querySelector('code')?.textContent || '';
+
+  try {
+    return JSON.parse(jsonStr);
+  } catch {
+    // Keep a single malformed schema from breaking the whole editor.
+    // Preserve the raw text so it can be repaired and re-saved.
+    return { error: 'Invalid JSON in schema.', invalid: jsonStr };
```
sharanyavinod: Is `invalid` key with a string value something that could be part of the schema itself?

kozmaadrian: Good catch! I've pushed a fix. I think a proper solution would require a larger refactoring, which is outside the scope of the issue we're addressing. We can improve the schema editor later, as the current implementation makes difficult to work with multiple schemas.

sharanyavinod (approved):

PR #645, opened 2026-08-06
feat: add new source.uploadMedia API

tripodsan:
## Summary
- Adds test coverage for the new `source.uploadMedia({ org, site, path, body })` API in `nx2/utils/api.js`: legacy delegation to `_saveDA` as FormData (including the stage `content.da.live` → `stage-content.da.live` contentUrl rewrite), hlx6 POSTs to the AEM media route with the correct `content-type` header, `contentUrl` prefix-normalization against the site's `aem.page` origin, non-ok passthrough for both branches, and the `/org/site/path` string call form.
- While writing the tests, found and fixed two bugs in the (previously uncommitted) `uploadMedia` implementation:
  1. The non-hlx6 branch fell through to also POST to the hlx6-only media endpoint whenever `DA_ADMIN` wasn't exactly `'https://stage-admin.da.live'` — i.e. for ordinary non-hlx6 sites in most environments, `uploadMedia` made an unintended second request after `_saveDA` had already completed. Fixed by returning unconditionally after the `_saveDA` call.
  2. In this repo's dev/test env `DA_ADMIN` is `'https://stage-admin.da.live'`, so the stage-content rewrite branch always runs for non-hlx6 uploads. When the returned `contentUrl`'s host wasn't `content.da.live` (no rewrite needed), the code had already consumed the response body via `resp.json()` and then returned that same drained `Response` — any caller calling `resp.json()` again would hit a "body stream already read" error. Fixed by always returning `adaptJsonResponse(resp, json)` in that branch.

## Test plan
- [x] `npm run test:file -- ./test/nx2/utils/api.test.js` — 120/120 passing (11 new tests for `uploadMedia`)
- [x] `npm test` (full suite) — all tests pass in isolation; the shared full-suite run is intermittently flaky under `--concurrent-browsers 4` on unrelated test files (pre-existing, not touched by this change)
- [x] `npx eslint nx2/utils/api.js test/nx2/utils/api.test.js` — clean

### DIFF
--- a/nx2/utils/api.js
+++ b/nx2/utils/api.js
@@ -82,7 +82,7 @@ export const config = {
       const resp = await daFetch({ url });
       if (resp.ok) {
         const cfg = object2sheet(await resp.json());
-        resp.json = () => cfg;
+        return adaptJsonResponse(resp, cfg);
       }
       return resp;
     }
@@ -299,9 +299,11 @@ export const source = {
   save: withArgs(async (opts) => {
     const { org, site } = opts;
     if (await isHlx6(org, site)) {
-      return this._saveHlx6(opts);
+      // eslint-disable-next-line no-underscore-dangle
+      return source._saveHlx6(opts);
     }
-    return this._saveDA(opts);
+    // eslint-disable-next-line no-underscore-dangle
+    return source._saveDA(opts);
   }),
 
   _saveHlx6: withArgs(async ({ org, site, path, body }) => {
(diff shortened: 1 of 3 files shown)


nx2/utils/api.js
```diff
@@ -296,31 +296,106 @@ export const source = {
...
+      if (resp.ok && DA_ADMIN === 'https://stage-admin.da.live') {
+        // special check for stage content. should be handled in stage da-admin ?
+        const json = await resp.json();
+        const sourceUrl = new URL(json.source.contentUrl);
+        if (sourceUrl.host === 'content.da.live') {
+          sourceUrl.host = 'stage-content.da.live';
+          json.source.contentUrl = sourceUrl.href;
+        }
+        return adaptJsonResponse(resp, json);
+      }
```
tripodsan: this is a workaround until we return the correct contentUrl when storing into stage from da-admin.
only needed when working against stage.

nx2/utils/api.js
```diff
@@ -700,12 +778,9 @@ function normalizePath(path) {
...
-// Shadow a Response's `json()` so it resolves to the DA-shaped
-// `{ source: { contentUrl } }`. Used for hlx6 saves, whose body is empty —
-// preserves the original ok/status/headers/permissions.
-function withSourceJson(resp, contentUrl) {
-  resp.json = async () => ({ source: { contentUrl } });
-  return resp;
+// Create a new response with a different JSON response body
+function adaptJsonResponse(resp, obj) {
+  return new Response(JSON.stringify(obj), resp);
 }
```
tripodsan, line 779: @anfibiacreativa better :-) ?

kptdobe (dismissed):

nx2/utils/api.js
kptdobe: https://github.com/adobe/da-admin/pull/312
Merged, now. Workaround should not be necessary anymore.

kptdobe (approved):

PR #652, opened 2026-08-10
fix resume conversation flow

sharanyavinod:
Ensure every resumed conversation also includes manifest id explicitly

https://da.live/canvas?nx=manifest&nxver=2#/exp-workspace/frescopa/index

### DIFF
--- a/nx2/blocks/chat/ao/chat-controller-ao.js
+++ b/nx2/blocks/chat/ao/chat-controller-ao.js
@@ -711,11 +711,39 @@ export default class ChatControllerAO {
     this._ws?.close();
   }
 
+  // Every interaction-response method below sends through this instead of a bare
+  // this._ws.send(...): AO's session for an episode goes idle almost immediately
+  // once a turn suspends, so the socket has commonly already silently reconnected
+  // by the time the user actually answers. Reconnect first if needed, same as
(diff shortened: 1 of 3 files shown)


nx2/blocks/chat/ao/chat-controller-ao.js
```diff
@@ -795,6 +876,8 @@ export default class ChatControllerAO {
         feedback,
         edited_plan_content: null,
       },
+      manifestId: AO_MANIFEST_ID,
+      debugMode: true,
```
hannessolo, line 880: Do we keep debugMode? What does it do?

sharanyavinod: We can pin to a specific manifest only if debugmode is set so in this case the `experience-workspace` manifest. IIUC most of the DA tools and capabilities are only available within this manifest.
On my list of things we would need to get clarity on, because I agree, debugmode and pinning to a manifest is not ideal to have on production especially if we want integration with coworker to behave seamlessly outside of a demo scenario.

hannessolo: ok, can you add it to the EW/AO gaps canvas?

hannessolo (approved):

PR #656, opened 2026-08-11
fix: lionbridge connector correctness/compliance fixes + test parity

markdaugherty:
## Summary
Continues the Lionbridge translation connector work from #651 (thanks @ravuthu). Fixes found by testing against the real Lionbridge Content API v2 and the real DA Translate app UI, plus compliance work against Lionbridge's dev guidelines, plus test parity with the other connectors (Trados/Smartling).

**Bug fixes (found via live API testing, not caught by mocked tests):**
- `providerId` must be sent on `submit`, not `createJob` — the API silently drops it on `POST /jobs`, then rejects `submit` with `"Job is missing providerId"`.
- `retrievefile` needs `Accept: application/octet-stream`, not the default `application/json` (403s otherwise). Other endpoints are unaffected.
- `auth.js` now resolves the da-etc origin via the shared `DA_ETC` export (env-override support via `?da-etc=`) instead of a hardcoded URL — needed to actually test this locally, and a real correctness fix regardless.

**Compliance with [Lionbridge's dev guidelines](https://developers.lionbridge.com/content/v2/docs/dev_guidelines.html):**
- Rate-limit handling: retry 429/503 with exponential backoff, honoring `Retry-After`.
- `jobName`/`requestName` truncated to the 250-byte limit.
- Generate and persist a per org/site/env connector GUID (in `.da/translate.json`, written once), prefixed onto `connectorName` per their connector-fingerprinting requirement.
- Call the `approve` endpoint after a successful download/save, closing out the request in Lionbridge's review workflow (`REVIEW_TRANSLATION` → `TRANSLATION_APPROVED`).
- No `cancelTranslation` export — their guidelines explicitly prohibit connectors from letting users cancel in-progress jobs; documented inline.

**Test parity:**
- Added `auth.test.js`, `connectorGuid.test.js`, and full `index.test.js` coverage for `sendAllLanguages`/`getStatusAll`/`saveItems`/`connect` (previously only `statusFor` had tests), matching Trados/Smartling's depth.
- Two tests directly regression-lock the `providerId` and `Accept`-header bugs above.

## Verification
- Standalone scripts against real Lionbridge staging APIs: full `sendAllLanguages` chain (createJob → upload → addRequest → submit), `getStatusAll`, `saveItems`/`retrievefile`, and `approve`.
[...]

PR #657, opened 2026-08-11
fix: retry Smartling API calls on 429/5xx to avoid batch-job failures

markdaugherty:
## Summary
The Smartling connector had no retry/backoff logic anywhere, so any `429` (rate limit) or transient `5xx` from Smartling's API failed the call outright. Two spots make this especially likely during larger batch jobs:

- `getStatusAll` polls `/file/progress` once per url in a tight sequential loop, no delay, no retry.
- `saveItems` downloads translated files 5-at-a-time via a concurrency-5 `Queue`.

Per [Smartling's own docs](https://help.smartling.com/hc/en-us/articles/1260805145550-Rate-Limits) (via search — the page itself sits behind a Cloudflare challenge), they enforce **two separate limits**: a request-rate limit and a concurrent-request limit, scoped per user/project/account depending on endpoint. Recommended handling is exponential backoff with jitter, capped at 30–60s, 5–10 max retries.

## Changes
- New shared `nx/blocks/loc/utils/fetchWithRetry.js`: exponential backoff + jitter, honors `Retry-After` when present, configurable `maxRetries`/`baseDelayMs`/`maxDelayMs`/`isRetryable` (defaults: 5 retries, 500ms base, 30s cap, retries on 429 and 5xx).
- Every `fetch` call in the Smartling connector now goes through it.
- Built as a shared utility rather than a one-off, since Lionbridge's connector (#656) needed the identical fix — that PR has been updated to use this same utility instead of its own inline copy, rather than maintain two copies.
- **Bug fix found during live verification** (`nx/blocks/loc/views/translate/translate.js`): `setupService()` built `this._service` via a spread/copy of `this.project.options.service`, taken once when the Translate view connects, before any job exists. Smartling's `sendAllLanguages` later sets `jobUid` directly on the original object; the copy never saw it, so `getStatusAll` crashed (`Cannot read properties of undefined (reading 'value')`) if you checked status in the same session without a page reload in between. Fixed to augment/reference the same object instead of copying it. Not Smartling-specific — affects any connector storing request state on `service` rather than per-lang.

## Tests
- New `test/loc/utils/fetchWithRetry.test.js`: retry-then-succeed on 429 (honoring `Retry-After`) and on 503 (exponential backoff), gives up after `maxRetries`, ignores non-retryable statuses, custom `isRetryable` override.
- Two new regression tests in the Smartling suite confirming `getStatusAll` and `saveItems` actually retry a 429 and succeed on the next attempt.
- Full `npm test` (1204 tests) and lint pass.

## Verification
Full end-to-end through the real DA Translate app UI (`da.live/apps/loc`) against a real test site (`scdemos/smartling-demo` — new GitHub repo + AEM Code Sync + DA content) and a real Smartling project (project-scoped API token, per Smartling's own recommendation for dev/testing): created a project, sent a real page for translation, checked status, and confirmed real machine-translated French content landed at `/fr/test-page.html` (not just an echo — genuinely translated by Smartling).

Also confirmed along the way that Smartling's connector must go through the `https://translate.da.live` proxy (never `api.smartling.com` directly from the browser — that hits a CORS block), which `resolveOrigin`'s legacy-rewrite logic already assumes but isn't otherwise documented anywhere.

Marking as draft pending final review pass.

### DIFF
--- a/nx/blocks/loc/connectors/smartling/index.js
+++ b/nx/blocks/loc/connectors/smartling/index.js
@@ -1,6 +1,7 @@
 import { Queue } from '../../../../../nx2/public/utils/tree.js';
 import { addDnt, removeDnt } from '../../dnt/dnt.js';
 import { DA_TRANSLATE } from '../../../../../nx2/utils/utils.js';
+import fetchWithRetry from '../../utils/fetchWithRetry.js';
 
 export const dnt = { addDnt };
 
@@ -46,7 +47,7 @@ function refreshTheToken(name, env, endpoint) {
     const body = JSON.stringify({ refreshToken: currRefreshToken });
     const opts = { ...BASE_OPTS, body };
 
-    const resp = await fetch(`${endpoint}/auth-api/v2/authenticate/refresh`, opts);
+    const resp = await fetchWithRetry(`${endpoint}/auth-api/v2/authenticate/refresh`, opts);
     if (!resp.ok) token = undefined;
     const json = await resp.json();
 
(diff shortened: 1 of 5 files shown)


markdaugherty:
Superseded by #662, which consolidates this with #661 into a single PR.

PR #659, opened 2026-08-12
feat: init ao chat

sharanyavinod:
- Includes basic chat using coworker
- extracts pills to be reusable
- enable selection context
- Common helpers for loading chat + events

Preview with new chat: https://ewaoinit--da-live--adobe.aem.live/canvas?nx=ewaoinit&nx-chat-ao=true#/exp-workspace/frescopa/coffee


nx2/blocks/chat-ao/ao-controller.js
```diff
@@ -0,0 +1,539 @@
...
+
+  _fetchEpisodeMessages(episodeId) { return fetchEpisodeMessages(episodeId); }
+
+  _fetchEpisodeContext(episodeId) { return fetchEpisodeContext(episodeId); }
+
+  _fetchWarmSession(episodeId) { return warmSession(episodeId); }
+
+  // Pre-warms the current episode's AO session while the user types. Existing
+  // episodes only, at most once per episode — see docs/chat-ao-component.md#session-warming.
+  async warmSession() {
```
anfibiacreativa, line 79: this one's outside the try/catch below it, so if warmSession's fetch fails we get an unhandled rejection on literally every keystroke. why don't we move the fetch inside the try too?

nx2/blocks/chat-ao/ao-controller.js
```diff
@@ -0,0 +1,539 @@
...
+      this._messages = [...this._messages, { role: 'assistant', content: `Error: ${err.message}` }];
+      this._done();
+    }
+  }
+
+  _handleServerEvent(evt) {
+    if (evt.type === AO_EVENT.SESSION_READY) {
+      const isNewEpisode = evt.episode_id && evt.episode_id !== this._episodeId;
+      this._episodeId = evt.episode_id ?? this._episodeId;
+      if (isNewEpisode) this._refreshEpisodeList();
```
anfibiacreativa: we're not awaiting or catching this call, so a failed fetch here just throws into the void. can we at least swallow it with a catch like the other network calls in this file?

nx2/blocks/chat-ao/question-card/question-card.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  _submit(e) {
+    e.preventDefault();
+    if (!this._canSubmit) return;
+    const answers = this.pending.questions.map((q) => ({
+      question_id: q.id,
+      selected_options: this._answerFor(q),
+    }));
+    this.onSubmit?.(answers);
```
anfibiacreativa, line 151: we use events everywhere else in this codebase, why is this component taking onSubmit/onDecline as plain function props instead of dispatching a submit/decline CustomEvent? it's not even declared in static properties so lit has no idea it exists

nx2/blocks/chat-ao/chat-ao.js
```diff
@@ -0,0 +1,491 @@
...
+      <div class="chat-form-wrap">
+        <nx-menu
+          class="slash-menu"
+          .ignoreFocus=${true}
+          .scoped=${true}
+          @select=${({ detail }) => this._onSlashSelect(detail.id)}
+          @mousedown=${(e) => e.preventDefault()}
+        ></nx-menu>
+        <nx-question-card
+          .pending=${this.pendingQuestion}
```
anfibiacreativa, line 419: same issue as question-card.js, we're wiring this up as callback props. should this just listen for @submit/@decline instead?

nx2/blocks/chat-ao/chat-ao.css
```diff
@@ -0,0 +1,711 @@
+:host {
```
anfibiacreativa, line 1: this is basically a fork of chat.css with new selectors sprinkled in. did we consider sharing the common message/input styling instead of copy pasting the whole sheet? this is going to be a nightmare to keep in sync

nx2/blocks/chat-ao/chat-ao.css
```diff
@@ -0,0 +1,711 @@
...
+      transform: rotate(180deg);
+    }
+
+    .tool-call-detail {
+      display: block;
+      padding-left: var(--s2-spacing-200);
+      color: var(--s2-gray-500);
+      overflow-wrap: break-word;
+      white-space: pre-wrap;
+      max-height: 120px;border-left: 2px solid var(--s2-gray-200);
```
anfibiacreativa: this line has two declarations crammed on one line, that's going to trip declaration-block-single-line-max-declarations. can we just split it?

nx2/blocks/chat-ao/utils/skills.js
```diff
@@ -0,0 +1,56 @@
+import { loadIms } from '../../../utils/ims.js';
+import { AO_MANIFEST_ID } from '../ao-constants.js';
+import { getOrgId, resolveAoHttpBase } from './uploads.js';
+
+const SKILLS_CACHE_KEY = `da-chat-ao-skills--${AO_MANIFEST_ID}`;
```
anfibiacreativa: I know there were conversations, but help me understand this better. we key the skills cache only by manifest id, but the response is scoped by x-tenant-id. switch orgs in the same browser and we'll render the previous org's skill list until the fetch resolves (or forever if it fails). should we fold org/tenant id into SKILLS_CACHE_KEY?

nx2/blocks/chat-ao/utils/uploads.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function base64ToBlob(base64, mediaType) {
+  const bytes = Uint8Array.from(atob(base64), (c) => c.charCodeAt(0));
+  return new Blob([bytes], { type: mediaType });
+}
+
+// POST /files/upload -> PUT the blob -> POST /finalize, per AO's Files API.
+export async function uploadAttachment({ fileName, mediaType, dataBase64 }) {
+  try {
```
anfibiacreativa, line 39: this exact header + base url build shows up three times now (here, skills.js, and as aoContext() in episodes.js). episodes.js already wrote the helper, why not export it from here and have skills.js and uploads.js call it instead of copy pasting?

nx2/blocks/shared/chat/voice-input.js
```diff
@@ -0,0 +1,90 @@
+// See docs/chat-ao-component.md#voice-input for the rationale behind this
```
anfibiacreativa, line 1: chat-ao.js is the only caller of this, chat/chat.js has no voice input at all.should we move it to blocks/chat-ao/utils/?

nx2/blocks/chat/chat.js
```diff
@@ -218,26 +133,19 @@ class NxChat extends LitElement {
       if (!this._explicitContext) this._applyContext(state);
     });
 
-    document.addEventListener(CHAT_EVENT.ADD_TO_CHAT, this._onAddToChat);
+    this._dnd = createFileDropHandlers({
```
anfibiacreativa, line 136: we pulled dnd.js and files.js out of chat.js but didn't add tests for either. slash-menu.js and voice-input.js got test files when they were extracted the same way, why not these two? the maxFiles capping and the silent catch-and-return-null path in buildAttachmentItems both deserve coverage

test/nx2/blocks/chat-ao/ao-controller.test.js
```diff
@@ -0,0 +1,1284 @@
...
+  const updates = [];
+  const sent = [];
+  const controller = new AoChatController({ onUpdate: (u) => updates.push(u) });
+  controller._ensureSocket = async () => { };
+  controller._fetchEpisodeContext = async () => null;
+  controller._ws = { send: (msg) => sent.push(JSON.parse(msg)) };
+  return { controller, updates, sent };
+}
+
+describe('ao-controller sendMessage', () => {
```
anfibiacreativa, line 21: sendMessage never gets exercised with attachments here, so uploadAttachment and the failed-uploads text building never run in this suite. can we add a case for a successful upload and one for a failed one? this is specially important for the .fig case (and overall, of course)

test/nx2/blocks/chat-ao/utils/uploads.test.js
```diff
@@ -0,0 +1,87 @@
+import { expect } from '@esm-bundle/chai';
+import {
+  getOrgId, resolveAoHttpBase, resolveAoWsBase,
+} from '../../../../../nx2/blocks/chat-ao/utils/uploads.js';
```
anfibiacreativa, line 4: uploadAttachment is the whole upload flow (initiate, PUT, finalize) with several failure branches, and none of it is covered here, only the url helpers are. can we mock fetch for each failure point plus the happy path?

test/nx2/blocks/shared/chat/slash-menu.test.js
```diff
@@ -0,0 +1,86 @@
+import { expect } from '@esm-bundle/chai';
+import { getSlashContext, spliceInput, syncSlashMenu } from '../../../../../nx2/blocks/shared/chat/slash-menu.js';
```
anfibiacreativa, line 2: createSlashMenu itself isn't even imported here, we only test the pure helpers. that's the piece wiring up onInput/onBlur/onKeydown against a real host, with real behavior like the debounced blur-close and keydown consuming arrow/enter/escape, and none of it is tested. can we add a fake host and cover it? (if not now, at some point?)

nx2/blocks/chat-ao/ao-controller.js
```diff
@@ -0,0 +1,537 @@
...
+  // See docs/chat-ao-component.md#connection-recovery — reattaches on any
+  // drop, mid-turn or idle, so cross-client updates keep arriving live; only
+  // a mid-turn failure is surfaced to the user.
+  async _recoverFromClose() {
+    if (!this._episodeId) return;
+    try {
+      await this._ensureSocket();
+      this._ws.send(JSON.stringify({ type: AO_FRAME.ATTACH }));
+    } catch (err) {
+      if (!this._thinking) return;
```
anfibiacreativa, line 303: this exact catch pattern shows up four times in this file. can we pull it into a _pushError(err) helper instead of copy pasting it every time?

nx2/blocks/chat-ao/ao-controller.js
```diff
@@ -0,0 +1,537 @@
...
+
+  async hydrateToolCall(toolCallId) {
+    const entry = this._messages.find((m) => m.toolCall?.toolCallId === toolCallId);
+    if (!entry || entry.toolCall.status !== 'summary') return;
+    if (entry.toolCall.calls || entry.toolCall.loadingCalls) return;
+    const { turnId } = entry.toolCall;
+
+    this._patchToolCall(toolCallId, { loadingCalls: true });
+
+    this._turnEventsCache ??= new Map();
```
anfibiacreativa, line 124: this cache never gets cleared, even on episode switch. is that intentional, or should we drop it in switchEpisode/startNewEpisode?

nx2/blocks/chat-ao/chat-ao.js
```diff
@@ -0,0 +1,491 @@
...
+
+const styles = await loadStyle(import.meta.url);
+const buttonStyle = await loadStyle(new URL('../../styles/buttons.css', import.meta.url).href);
+const artifactStyle = await loadStyle(new URL('./artifacts/artifacts.css', import.meta.url).href);
+
+const { codeBase } = getConfig();
+
+const icon = (name) => html`<svg viewBox="0 0 20 20" aria-hidden="true"><use href="${codeBase}/img/icons/${ICON_NAMES[name]}.svg#icon"></use></svg>`;
+
+function isAllowedFile(file) {
```
anfibiacreativa, line 50: this is a pure extension check, why isn't it in nx2/utils/ so other blocks can reuse it instead of redefining it per component?

nx2/blocks/chat-ao/chat-ao.css
```diff
@@ -0,0 +1,711 @@
...
+    display: none;
+  }
+
+  button {
+    border: none;
+    padding: 0;
+    cursor: pointer;
+    font-family: inherit;
+
+    &.action-btn {
```
anfibiacreativa: .action-btn and it's not used anywhere, only nx-action-btn-*. can we just delete this dead selector?

nx2/blocks/chat-ao/chat-ao.js
```diff
@@ -0,0 +1,491 @@
...
+  render() {
+    const { view } = this._context ?? {};
+    const prompts = (this._prompts ?? [])
+      .filter((p) => !p.area || p.area === 'all' || p.area === view);
+
+    return html`
+      <nx-popover class="prompts-popover">
+        <nx-prompts
+          .prompts=${prompts}
+          .onSend=${(p) => this._sendPrompt(p)}
```
anfibiacreativa, line 351: could we convert to event instead of prop?

nx2/blocks/chat-ao/chat-ao.css
mhaack: fixed

nx2/blocks/chat-ao/ao-controller.js
mhaack: fixed

nx2/blocks/chat-ao/ao-controller.js
mhaack: fixed

nx2/blocks/chat-ao/chat-ao.css
mhaack: fixed

nx2/blocks/chat-ao/utils/skills.js
mhaack: fixed

nx2/blocks/chat-ao/question-card/question-card.js
sharanyavinod: 1. Not true, we do use events everywhere we need communication to be generic and there are multiple consumers. But in cases like this where there's one clear parent/consumer, we already use props, and introducing custom events here would be overkill imo.
2. onSubmit doesn't need to be a static property. There's no reason for Lit to manage it as a reactive property here and we don't want the component to react to onSubmit changing - it's just a callback we invoke.

nx2/blocks/chat-ao/chat-ao.css
sharanyavinod: I did, but its more painful keeping variations of selectors and deciding which is relevant and which is not, considering we will need to deprecated nx-chat version soon anyway. Ideally we dont touch chat.css at all anymore and any new changes would only be here.

PR #661, opened 2026-08-12
feat: add auto-authorization config option for Smartling

markdaugherty:
## Summary
Smartling's `job-batches-api` already supports an `authorize` flag on batch creation — previously hardcoded to `false` in this connector, meaning every job requires manual authorization in Smartling's own dashboard before translation work starts.

Adds a new site config value, `translation.service.{env}.autoAuthorize` ("yes" to enable), read at send time and passed through to `createBatch`. When enabled, a job's content is immediately authorized for translation as soon as its batch finishes processing.

## Design choice
Site-level config rather than a per-project toggle in the Options UI: the existing custom-fields mechanism (`translation.service.custom.*`, used for Trados's Template/Notes fields) only supports dropdowns and free-text, not checkboxes, and this reads more like a fixed site-wide policy (does this org want translation to start immediately, or always go through manual review?) than something to decide per project.

## Tests
- Three new tests: default (`authorize: false`), enabled (`autoAuthorize: 'yes'` → `authorize: true`), and explicitly disabled (`autoAuthorize: 'no'` → `authorize: false`, confirming it's an exact-match check, not a truthy check on any non-empty string).
- Full `npm test` and lint pass.

## Note on branch base
Stacked on top of #657 (the Smartling rate-limit fix) rather than `main` directly, since this connector needed that fix regardless. The diff will show just this change once #657 merges.

### DIFF
--- a/nx/blocks/loc/connectors/smartling/index.js
+++ b/nx/blocks/loc/connectors/smartling/index.js
@@ -1,6 +1,7 @@
 import { Queue } from '../../../../../nx2/public/utils/tree.js';
 import { addDnt, removeDnt } from '../../dnt/dnt.js';
 import { DA_TRANSLATE } from '../../../../../nx2/utils/utils.js';
+import fetchWithRetry from '../../utils/fetchWithRetry.js';
 
 export const dnt = { addDnt };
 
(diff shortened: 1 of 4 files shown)


markdaugherty:
Superseded by #662, which consolidates this with #657 into a single PR.

PR #666, opened 2026-08-13
fix(ew-actions): when editor.hide publish is set, also hide publish for canvas

### LINKED ISSUE #664: [ew] support `editor.hidePublish`
DA Edit hides the publish button via a config. Canvas needs to support this too, which uses `ew-actions`

shsteimer:
Mirrors da-live's da-title.js filterActions(): fetches org/site DA config docs on hash change, ORs any editor.hidePublish rows, and drops publish from the Send button's menu items when the current document's fullpath starts with a configured prefix.

Please always provide the [GitHub issue(s)](https://github.com/adobe/da-nx/issues) your PR is for, as well as test URLs where your change can be observed (before and after):

Fix https://github.com/adobe/da-nx/issues/664

https://feature-ew-actions-hide-publish--da-nx--adobe.aem.live/

should merge with https://github.com/adobe/da-live/pull/1245


sharanyavinod (approved):

nx2/blocks/ew-actions/ew-actions.js
```diff
@@ -97,6 +99,25 @@ class NXEwActions extends LitElement {
     this._unsubHash?.();
   }
 
+  update(changed) {
+    super.update(changed);
+    if (changed.has('_hashState') && this._hashState) this._filterHidePublish();
```
anfibiacreativa: _filterHidePublish is fired from update() without a catch, so a config fetch failure is an unhandled rejection. please wrap it and default _hidePublish to false on error

nx2/blocks/ew-actions/ew-actions.js
```diff
@@ -97,6 +99,25 @@ class NXEwActions extends LitElement {
...
+    const fullpath = buildPrepareDetails(this._hashState)?.fullpath;
+    if (!org || !site || !fullpath) {
+      this._hidePublish = false;
+      return;
+    }
+
+    const configs = await Promise.all(fetchDaConfigs({ org, site }));
+    const configTab = configs.flatMap((config) => getFirstSheet(config) || []);
+    const publishConfigs = configTab.filter((c) => c.key === 'editor.hidePublish');
+    this._hidePublish = publishConfigs.some((c) => fullpath.startsWith(c.value));
```
anfibiacreativa: should we guard against an empty editor.hidePublish value? right now a blank row hides publish everywhere via startsWith('').

anfibiacreativa: nit: _shouldHidePublish reads truer to a boolean (it mirrors the config key, so up to you).

anfibiacreativa (dismissed):

shsteimer:
@anfibiacreativa appreciate the review.

Agree that changing name from filter->should is better, but I don't like `_shouldHidePublish` as a function with a side-effect. Obviously a matter of opinion, but I prefer `should` functions to return a boolean value and let the caller assign/handle as needed. At least to my mind that makes more sense. To that end, I extracted `shouldHidePublish` as a module level function that I'm now calling from inside of `_updateHidePublish`. Hope that works for you.

Other updates made to handle errors and empty values.

mhaack (approved):

PR #668, opened 2026-08-14
feat(form): clearer validation, field help text, and array add/remove UX

kozmaadrian:
## Summary

Improves the Structured Content **form editor**: validation errors now surface clearly on the right controls, schema `description`s render as field help text, and array add/remove behaves predictably. Also bumps `@adobe/da-sc-sdk` to **0.4.0** (validation hardening + the `description` annotation) and re-vendors the bundle.

Pairs with da-sc-sdk [#14](https://github.com/adobe/da-sc-sdk/pull/14) (released as 0.4.0).

## What changed

**Validation feedback**
- Errors render where the author expects them: leaf errors inside the field, array/object **group-level** errors on the group, and nested-child errors at their own pointer.
- Required fields are marked with a red asterisk.

**Field help text (schema `description`)**
- Every field component (input, textarea, checkbox, number, picker) gained a `description` prop and renders `label → input → description → error`.
- Migrated off the old bottom description slot; the schema selector's "Schema Editor" guidance moved to a sibling hint.
- Spectrum behavior: while a field is invalid, the error replaces the help text.

**Array add/remove UX**
- Required arrays seed empty rows up to `minItems` on load (UI-only convenience; empty rows prune on save, so headless/CLI consumers still get none).
- Removal is never floored — any row can be deleted; dropping below `minItems` (or emptying a required array) is surfaced as a validation error, not blocked. This matches the engine's non-blocking model and the 0.4.0 `canRemove`.
- Add is capped at `maxItems` on the raw row count.

**Tests**
- Form UI: editor loads, model↔DOM correspondence, add/remove/insert/reorder array actions end-to-end, validation-error rendering (leaf / array / object / nested), descriptions, required asterisk, and remove-gating.
- SDK↔form contract tests against the real bundle, so a future re-vendor that drifts the API fails loudly.

**Test url**
- https://da.live/form?nx=vldtnfix#/exp-workspace/frescopa/forms/validation-showcase


nx/blocks/form/fields/checkbox.js
```diff
@@ -39,6 +41,7 @@ class FormCheckbox extends LitElement {
         <span class="form-switch-label"><slot></slot></span>
       </label>
       ${this.error ? html`<p class="form-field-error">${this.error}</p>` : nothing}
+      ${showHint ? html`<p class="form-field-description">${this.description}</p>` : nothing}
```
sharanyavinod, line 44: Not a blocker for this PR, but might be nice to look into -> I see a lot of duplication across the fields related to how hints/descriptions/error etc are handled. May be interesting to explore if this can be simplified?

nx/blocks/form/views/editor.css
```diff
@@ -209,6 +209,23 @@
   margin-left: 2px;
 }
 
+.editor-root .form-node-error {
+  /* Group-level validation message (required / minItems / maxItems on an
+     object or array). Mirrors the primitive field's .form-field-error, whose
+     distance from its control is the .form-field flex gap (--s2-spacing-75);
+     match that gap here so group and input errors share the same rhythm. */
+  margin: var(--s2-spacing-75) 0;
+  font-size: var(--s2-body-size-xs, 12px);
```
sharanyavinod: font size should use the --s2-component-s-regular-font-size pattern as much as possible in such cases, body-size is meant to be a more generic fallback iiuc. This is a recurring pattern across the repo we need to improve however, but would be good to fix already in new code :)

nx/blocks/form/editor.css
```diff
@@ -112,6 +112,22 @@ nx-sidebar {
...
+     editor's fields use between input and description (.form-field). The
+     form's own 16px gap then separates this group from the Create button. */
+  display: flex;
+  flex-direction: column;
+  gap: var(--s2-spacing-75);
+}
+
+.nx-form-schema-form .nx-form-schema-hint {
+  /* Match the field/group help text in the form editor (.form-field-description
+     / .form-node-description): small and muted, sitting under the picker. */
```
sharanyavinod: Quite a few of the comments seem to be unnecessary - lets cleanup if possible or make it more concise

nx/blocks/form/fields/checkbox.js
kozmaadrian: Yeah, good catch. I tried pulling the shared bits into a "field shell" helper so the fields didn't repeat the label/wrap/icon/footer markup.

But I backed it out. It ended up needing an 8-param helper, and the fields are different enough (checkbox is a switch, picker has a chevron, number has steppers) that it didn't really help. It mostly just made you jump between two files to see what one field renders.

Left them self-contained for now. Happy to revisit if we add more fields.

nx/blocks/form/views/editor.css
kozmaadrian: Done, switched to var(--s2-component-s-regular-font-size, 12px)

nx/blocks/form/editor.css
kozmaadrian: Done, dropped the comments that just restated the CSS.

mhaack (approved):
Looks ok to me

PR #670, opened 2026-08-15
fix(quick-edit): pass document to loadPage, so a site that declares the argument decorates

benpeter:
`setBody` called the site's `loadPage` with no argument, so a site that declares one got `undefined` and threw instead of re-decorating.


benpeter:
aem.live trips over this when quick-edit triggers the re-decoration

```
scripts.js:985 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'querySelector')
    at loadEager (scripts.js:985:20)
    at async Object.loadPage (scripts.js:1067:3)
    at async setBody (quick-edit.js:29:3)
```

mhaack (approved):

PR #674, opened 2026-08-18
feat: opt-in image translation for GLaaS multimodal send, fix 5 MB limit

janaki-r-bhagwath:
opt-in image translation for GLaaS multimodal send
- Add a DA_SDK panel (plugin/) for authors to mark individual images for translation, stored in the page's da-metadata (loc-images)
- Broaden multimodal image eligibility beyond content.da.live to any absolute http(s) image
- Protect da-metadata from translation wrapping; strip loc-images marks on translate save-back only, not on sync/rollout
- Fix DNT's image relativization silently dropping marked AEM-hosted images from collection and save-back rewriting
- Fix CORS-blocked image fetches (e.g. published .aem.live media) via the existing da-etc proxy
- Fix plugin thumbnails for content.da.live/admin.da.live (authenticated
  fetch) and aem.page (preview.da.live cookie) images
  
  fix 5 MB limit - replace media-bus image upload with DA source save for GLaaS translations
- Media-bus upload capped at an observed 5MB; DA /source allows 20MB
- Helix resolves content.da.live refs to public media_ URLs on preview/publish
- Path-based for now; dedup across shared images deferred

### DIFF
--- a/nx/blocks/loc/connectors/glaas/imageSelections.js
+++ b/nx/blocks/loc/connectors/glaas/imageSelections.js
@@ -5,6 +5,13 @@ export const LOC_IMAGES_KEY = 'loc-images';
 export const DA_METADATA_SELECTOR = 'body > .da-metadata';
 const ELIGIBLE_IMAGE_EXTS = new Set(['png', 'jpg', 'jpeg']);
 const AEM_PAGE_SUFFIX = '.aem.page';
+const ELIGIBLE_IMAGE_HOSTS = new Set(['content.da.live']);
+const ELIGIBLE_IMAGE_HOST_SUFFIXES = ['.aem.live', '.aem.page'];
+
+function isEligibleImageHost(hostname) {
+  return ELIGIBLE_IMAGE_HOSTS.has(hostname)
+    || ELIGIBLE_IMAGE_HOST_SUFFIXES.some((suffix) => hostname.endsWith(suffix));
+}
 
 export function toHref(src) {
   try {
@@ -34,6 +41,7 @@ export function isEligibleMultimodalImageUrl(src) {
   try {
     const url = new URL(src);
     if (url.protocol !== 'http:' && url.protocol !== 'https:') return false;
+    if (!isEligibleImageHost(url.hostname)) return false;
     const pathname = decodeURIComponent(url.pathname);
     const filename = pathname.split('/').pop() ?? '';
     const dot = filename.lastIndexOf('.');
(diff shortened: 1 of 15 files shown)


nx/blocks/loc/connectors/glaas/plugin/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+// DA-internal storage requires the same Bearer token the plugin already uses for the page
+// itself - an <img> tag can't send one, so these need fetchAuthenticatedThumbnail below
+// instead of being usable as a thumbnail src directly.
+const DA_INTERNAL_HOSTS = new Set(['content.da.live', 'admin.da.live']);
+function needsAuthenticatedFetch(href) {
+  try {
+    return DA_INTERNAL_HOSTS.has(new URL(href).hostname);
+  } catch {
+    return true;
```
jenssingler: I think this should be `false`. Otherwise this risks to leak the token (if there is a HTML with a manipulated `img` href).

nx/blocks/loc/connectors/glaas/imageSelections.js
```diff
@@ -0,0 +1,162 @@
...
+export function aemPageToPreviewDaLiveUrl(imageUrl, { ref, repo, org }) {
+  const url = new URL(imageUrl);
+  return `${getLivePreviewUrl(org, repo, ref)}${url.pathname}${url.search}`;
+}
+
+// Memoized per {org}/{repo}/{ref} so concurrent callers share one cookie exchange.
+const livePreviewLogins = new Map();
+export function ensureLivePreviewLogin({ org, repo, ref }) {
+  const key = `${org}/${repo}/${ref}`;
+  if (!livePreviewLogins.has(key)) livePreviewLogins.set(key, livePreviewLogin(org, repo, ref));
```
jenssingler: I think that you should handle that differently, since the Map will stay the whole browser session and one failed attempt will mark the host as `false` until you reload.
Maybe this could work:
```
export function ensureLivePreviewLogin({ org, repo, ref }) {
  const key = `${org}/${repo}/${ref}`;
  let login = livePreviewLogins.get(key);
  if (!login) {
    login = livePreviewLogin(org, repo, ref).then((ok) => {
      if (!ok) livePreviewLogins.delete(key);
      return ok;
    });
    livePreviewLogins.set(key, login);
  }
  return login;
}
```
And then in fetchMultimodalImage you could do:
```
  const loggedIn = await ensureLivePreviewLogin(aemPageHost);
  if (!loggedIn) {
    return { error: 'Error fetching image.', step: `fetch-image-${imageIndex}` };
  }
```

jenssingler (dismissed):
I'm approving since I'm not sure of the two things I've found are critical enough to block the PR.

janaki-r-bhagwath:
@jenssingler , addressed the review comments. Can you please review again.

jenssingler (approved):

PR #680, opened 2026-08-20
feat(form): native date, time and date-time fields

kozmaadrian:
## TODO

- [ ] **Release the `da-sc-sdk` with date/time format validation** [PR](https://github.com/adobe/da-sc-sdk/pull/15) and re-vendor the published version — this PR currently vendors a dev build of the SDK in `nx/deps/da-sc-sdk/dist/`. The SDK source changes (format extraction, value-shape validation, docs) live on a separate `da-sc-sdk` branch.

---

## Summary

Adds native **date**, **time**, and **date-time** field support to the SC form editor, driven entirely by the JSON Schema `format` keyword. A single `form-date` component wraps the browser's native `<input>` (`date` / `time` / `datetime-local`), so authors get real pickers and keyboard entry — including on mobile — instead of a plain text box.

**Test URL:** https://da.live/form?nx=formdate#/exp-workspace/frescopa/forms/date-input-test

**Before**: the time, date, and date-time inputs are simply text inputs.

Screenshot 2026-08-20 at 11 00 40 am

**After**: the time, date, and date-time fields use native HTML input elements.

format: date (floating date)
Screenshot 2026-08-20 at 10 59 58 am

format: date-time (the UI displays the users local time, however saves the same time in UTC always)
Screenshot 2026-08-20 at 11 00 08 am

format: time (float time, no timezone)
Screenshot 2026-08-20 at 11 00 18 am

## How it works

The editor maps the schema `format` to one parameterized component:

| schema `format` | native input | stored value |
| --- | --- | --- |
| `date` | `date` | `YYYY-MM-DD` (floating, no timezone) |
| `time` | `time` | `HH:MM` (floating, no timezone) |
| `date-time` | `datetime-local` | `YYYY-MM-DDTHH:MM:00Z` (absolute instant, UTC) |

- **`date` / `time` are floating** — they mean the same everywhere (a publish date, store opening hours).
- **`date-time` is store-UTC / display-local** — the picker shows and edits the value in the viewer's local zone and converts back to UTC on save, so the instant stays stable no matter which zone an author edits from.

## Validation

Validation is owned by the SDK (the trust boundary — a script writing directly can't bypass it), not the widget:

- `date` / `time` are checked for real calendar/clock validity (`2026-02-30`, `24:00` are rejected).
- `date-time` is **deliberately stricter than RFC 3339**: only UTC `Z` with zero seconds is accepted, so the editor and any SDK-only writer produce one identical, comparable value (offsets, non-zero seconds, and sub-minute precision are rejected).
- Empty optional fields are treated as absent — `format` is not enforced on an empty value.

## Edge cases

- Pre-1900 dates carry sub-minute **Local Mean Time** offsets; the local→UTC conversion rounds to the minute so those can't leak seconds and break the canonical form.
- Early-CE years are zero-padded to 4 digits; the native year field is capped to 4 digits (some browsers otherwise allow 6).

## Compatibility

- **Additive** for any schema that doesn't use these formats — nothing changes.
- Schemas that **already** declare `format: date | time | date-time` now render as pickers **and** get validated. Existing documents whose values aren't canonical (especially date-times with an offset or non-zero seconds) would surface a validation error, so it's worth a quick audit of existing date-time data. Validation is not gated on save, so this surfaces an error rather than blocking.

## Testing

- Unit tests for the widget, the local↔UTC helpers, and the editor dispatch, verified across time zones (incl. Europe/Zurich and America/New_York, which exercise the LMT edge). Full suite green.


nx/blocks/form/fields/date.css
```diff
@@ -0,0 +1,42 @@
...
+  box-sizing: border-box;
+  width: 100%;
+  height: 32px;
+  padding: 0 calc(var(--s2-spacing-200) - 2px);
+  border: 2px solid var(--s2-gray-300);
+  border-radius: var(--s2-corner-radius-500);
+  background: var(--s2-gray-25);
+  color: var(--s2-gray-800);
+  font-family: inherit;
+  font-size: var(--s2-body-size-s, 14px);
```
sharanyavinod, line 11: Please avoid using body-size for font - ideal is to use --s2-component-m-regular-font-size and so on depending on requirement.
Also for generic input/error styles, would be nice if you can reuse https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html#form - ignore if its inconsistent with the remaining fields in sc atm. Could be a follow up task for later to consolidate.

nx/blocks/form/fields/date.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  _onInput(e) {
+    const raw = e.target.value;
+    const next = this.type === 'datetime' ? localToUtc(raw) : raw;
+    this._lastValue = next;
+    this.value = next;
+    this.dispatchEvent(new Event('input', { bubbles: true, composed: true }));
+    this.dispatchEvent(new Event('change', { bubbles: true, composed: true }));
```
sharanyavinod: Why do we fire both together always? In this way any consumer would never be able to differentitate between when an input happened vs when a change happened.

nx/blocks/form/fields/date.js
```diff
@@ -0,0 +1,86 @@
...
+  render() {
+    const showHint = !this.error && this.description;
+    return html`
+      <div class="form-field${this.error ? ' has-error' : ''}">
+        ${this.label ? html`<label for="form-date">${this.label}${this.required ? html`<span class="form-required">*</span>` : nothing}</label>` : nothing}
+        <input
+          id="form-date"
+          type=${NATIVE_TYPE[this.type] ?? 'date'}
+          min=${RANGE[this.type]?.min ?? nothing}
+          max=${RANGE[this.type]?.max ?? nothing}
```
sharanyavinod, line 70: If this.required is available why not provide the attribute to input as well?

nx/blocks/form/views/editor.js
```diff
@@ -252,6 +253,23 @@ class Editor extends LitElement {
...
+      return html`
+        <form-date
+          data-pointer=${pointer}
+          .type=${widgetType}
+          .label=${label}
+          .required=${showRequired}
+          .error=${error}
+          .description=${description}
+          .value=${value ?? ''}
+          ?disabled=${readonly}
```
sharanyavinod, line 267: If the user is setting something as readonly, it should be passed down as readonly to the input fields as well. Disabled has a very different behaviour from readonly, and user expectation would be aligned to web standard behaviour.

nx/blocks/form/fields/date.js
kozmaadrian: Good catch. It now emits the same event type as the native event instead of firing both, matching `form-input`, so input and change stay distinct.

nx/blocks/form/fields/date.css
kozmaadrian: These match the other fields today (`input`, `number` all use `--s2-body-size-s`), so I kept date aligned with them. I will move everything to the component tokens in a separate consolidation task, as you suggested.

nx/blocks/form/fields/date.js
kozmaadrian: Same as the other fields today: `form-input` and `form-number` show the asterisk on the label but don't set required on the control. I will address it across all fields in the same follow up, likely via an aria attribute since validation is SDK driven.

nx/blocks/form/views/editor.js
kozmaadrian: Agreed they differ. The whole editor maps readonly to disabled on every field, so I would rather fix it consistently across all fields than diverge here. Noting the value is still saved when disabled (data comes from the model, not a form submit), so the benefit is mainly letting the user focus and copy.

nx/blocks/form/fields/date.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  _onEvent(e) {
+    const raw = e.target.value;
+    const next = this.type === 'datetime' ? localToUtc(raw) : raw;
+    this._lastValue = next;
+    this.value = next;
+    e.stopPropagation();
+    this.dispatchEvent(new Event(e.type, { bubbles: true, composed: true }));
```
sharanyavinod: This still doesn't feel like the right approach to me. IMO consumer should receive the events as close as possible to how it was natively fired. So:
-  `input` event should be left to propagate on its own without any intervention since it can cross shadow dom boundaries
- `change` gets fired explicitly since it needs update to composed to cross the boundary, but while trying to keep intact the event structure as much as possible instead of a default Event.

kozmaadrian: Very good point, thanks. Went with just letting the input event propagate on its own now, no intervention. It's composed so it already crosses the shadow boundary, and the editor only listens to input anyway. Dropped the extra dispatch since nothing actually consumes change today. Can always add a composed change later if something needs it.

sharanyavinod:
Looks great, could you please create follow up issues to track required and disabled/readonly concerns across all fields?

sharanyavinod (dismissed):

tyge68 (approved):

mhaack (approved):
LGTM

